PR #29, opened 2015-08-27
Updated All NPC Format to VS Standard

Crypting:
This also includes moving Greetings that were in the wrong spot.

### DIFF
--- a/system/scripts/npcs/dunbarton/walter.cs
+++ b/system/scripts/npcs/dunbarton/walter.cs
@@ -34,8 +34,8 @@ public class WalterScript : NpcScript
 		AddPhrase("What do you need?");
 	}
 
-    protected override async Task Talk()
-    {
+	protected override async Task Talk()
+	{
 		SetBgm("NPC_Walter.mp3");
 
 		await Intro(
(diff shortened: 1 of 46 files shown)


system/scripts/npcs/dunbarton/walter.cs
```diff
@@ -62,15 +62,69 @@ public override void Load()
 				return;
 
 			case "@repair":
-				Msg("Repair? What is it that you want to repair? Let's have a look.<br/>I can take care of general goods like instruments, glasses, and tools.<br/>My skills are not what they used to be, so I won't charge you a lot...");
-				Msg("Unimplemented");
+				Msg("Repair? What is it that you want to repair? Let's have a look.<br/>I can take care of general goods like instruments, glasses, and tools.<br/>My skills are not what they used to be, so I won't charge you a lot...<repair rate='92' stringid='(*/misc_repairable/*)|(*/cashchair/*)' />");
```
exectails: This would've been better in a separate commit.

Crypting: Oh oops, I could have swore this was already pulled... Sorry bout that.

PR #30, opened 2015-08-27
Update NpcScript.cs

Blade3575:
Implemented official repair rate when item is blessed.


exectails:
Is there an official reference for this?

Blade3575:
Only if you take the word of a person who has done extensive research.

exectails:
Oh well, better than the 1% guess^^

PR #40, opened 2015-09-15
Dice Tossing

UnknownVoid:
Working Dice Tossing Skill.
Able to learn Dice tossing when Equipped.

Todo
Ammo Handling.

### DIFF
--- a/src/ChannelServer/Skills/Actions/DiceTossing.cs
+++ b/src/ChannelServer/Skills/Actions/DiceTossing.cs
@@ -52,7 +52,7 @@ namespace Aura.Channel.Skills.Action
 			}
 
 			// there seems to be client side code to prevent this, but just incase..
-			if (creature.RightHand != null && !creature.RightHand.HasTag("/dice/"))
+			if (creature.RightHand == null || !creature.RightHand.HasTag("/dice/"))
 			{
 				Send.Notice(creature, Localization.Get("You must equip one Six Sided Dice to use this Action."));
 				return false;
@@ -96,6 +96,14 @@ namespace Aura.Channel.Skills.Action
 			var unkInt3 = packet.GetInt();
 			var unkInt4 = packet.GetInt();
 
+			var areaPosition = new Position(location);
+
+			// Check range
+			if (!creature.GetPosition().InRange(areaPosition, Range))
+			{
+				Send.Notice(creature, Localization.Get("Out of range."));
+			}
+
 			// Reduce Dice
 			if (creature.Inventory.RightHand != null)
 				creature.Inventory.Decrement(creature.Inventory.RightHand);
@@ -117,14 +125,6 @@ namespace Aura.Channel.Skills.Action
 			var location = packet.GetLong();
 			var unkInt3 = packet.GetInt();
 			var unkInt4 = packet.GetInt();
-
-			var areaPosition = new Position(location);
-
-			// Check range
-			if (!creature.GetPosition().InRange(areaPosition, Range))
-			{
-				Send.Notice(creature, Localization.Get("Out of range."));
-			}
 
 			Send.SkillComplete(creature, skill.Info.Id, location, unkInt3, unkInt4);
 		}
--- a/src/ChannelServer/Skills/Actions/DiceTossing.cs
+++ b/src/ChannelServer/Skills/Actions/DiceTossing.cs
@@ -45,9 +45,16 @@ namespace Aura.Channel.Skills.Action
 		/// <returns></returns>
 		public bool Prepare(Creature creature, Skill skill, Packet packet)
 		{
-			if (creature.Inventory.Count(ItemId) == 0)
+			if (!creature.Inventory.Has(ItemId))
 			{
-				Send.Notice(creature, Localization.Get("You need at least one Six Sided Dice."));
+				Send.Notice(creature, Localization.Get("You need at least one Six Sided Dice"));
+				return false;
+			}
+
+			// there seems to be client side code to prevent this, but just incase..
+			if (creature.RightHand != null && !creature.RightHand.HasTag("/dice/"))
+			{
+				Send.Notice(creature, Localization.Get("You must equip one Six Sided Dice to use this Action."));
 				return false;
 			}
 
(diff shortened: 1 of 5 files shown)


src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// <summary>
+		/// Prepares skill, fails if no Dice is found.
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		/// <returns></returns>
+		public bool Prepare(Creature creature, Skill skill, Packet packet)
+		{
+			if (creature.Inventory.Count(ItemId) == 0)
```
exectails: `CreatureInventory.Has` might be more appropriate for this.

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// Readies the skill.
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		/// <returns></returns>
+		public bool Ready(Creature creature, Skill skill, Packet packet)
+		{
+			//skill.Stacks = 1; //I will leave this here.
+			Send.SkillStackSet(creature, skill.Info.Id, 1);
```
exectails: Why don't you set it properly? Sending just the packet is completely pointless, actually, you put the client and the server out of sync doing that.

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Use(Creature creature, Skill skill, Packet packet)
+		{
+			var location = packet.GetLong();
+			var unkInt3 = packet.GetInt();
+			var unkInt4 = packet.GetInt();
+
+			// Send.ItemAmount(creature, ); Remove 1 item
```
exectails: Come on, it's one line that you can copy almost 1:1 from ranged, just changing the item to reduce. You can do it!

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+			Send.UseMotion(creature, 27, 2, false, false);
+			Send.Effect(creature, Effect.Dice, "process", location, (byte)3);
+			Send.SkillUse(creature, skill.Info.Id, location, unkInt3, unkInt4);
+			skill.Stacks = 0;
```
exectails: So here you do set Stacks? Setting Stacks sends a packet btw, you're sending two stack packets to the client.

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Complete(Creature creature, Skill skill, Packet packet)
+		{
+			var location = packet.GetLong();
+			var unkInt3 = packet.GetInt();
+			var unkInt4 = packet.GetInt();
+
+			// Check range
+			if (!creature.GetPosition().InRange(creature.GetPosition(), Range))
```
exectails: `if (creature is not in range of creature)`...? That's unlikely to happen^^"

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+			{
+				Send.Notice(creature, Localization.Get("Out of range."));
+				goto L_End;
+			}
+
+			L_End:
```
exectails: Is there any reason for this label's existence?

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,137 @@
...
+		}
+
+		/// <summary>
+		/// Cancels skill, Cancels Motion and returns the character to idle position.
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		public void Cancel(Creature creature, Skill skill)
+		{
+			skill.Stacks = 0;
```
exectails: Same Stacks problem.

src/ChannelServer/Skills/Actions/DiceTossing.cs
UnknownVoid: yeah, i tested with just one packet for don't know why but dice skill hangs on complete after skill use is executed, so added skill.Stacks = 0; to fix this issue, This results in 2 packets. I'm looking into why the skill hangs on complete.

src/ChannelServer/Skills/Actions/DiceTossing.cs
UnknownVoid: i think, it was suppose to be location position not the creature ._.; fixed.

src/ChannelServer/Skills/Actions/DiceTossing.cs
UnknownVoid: Nope, Removed it.

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,144 @@
...
+			var location = packet.GetLong();
+			var unkInt3 = packet.GetInt();
+			var unkInt4 = packet.GetInt();
+
+			var areaPosition = new Position(location);
+
+			// Check range
+			if (!creature.GetPosition().InRange(areaPosition, Range))
+			{
+				Send.Notice(creature, Localization.Get("Out of range."));
```
exectails: Sending this notice at Complete is a little pointless, because at this point the skill is already done. You should do it in Use, and possibly stop the skill if it's out of range, because if it is, the player might be cheating. For canceling skills midway, there are multiple "Silent" cancel packets, like `SkillUseSilentCancel`.

src/ChannelServer/Skills/Actions/DiceTossing.cs
```diff
@@ -0,0 +1,144 @@
...
+		public bool Prepare(Creature creature, Skill skill, Packet packet)
+		{
+			if (!creature.Inventory.Has(ItemId))
+			{
+				Send.Notice(creature, Localization.Get("You need at least one Six Sided Dice"));
+				return false;
+			}
+
+			// there seems to be client side code to prevent this, but just incase..
+			if (creature.RightHand != null && !creature.RightHand.HasTag("/dice/"))
```
exectails: This should be `rightHand == null || !rightHand.HasTag`. With the check written like you have it, someone could use the skill without equipped dice, because your check doesn't fail if RightHand is null (empty).

Doing such checks is definitely a good thing, even with client checks in place.

PR #41, opened 2015-09-15
Bhafel Huntress Arrow Exception

xeroplz:
Bhafel Huntess does not get removed on skill use.

### DIFF
--- a/src/ChannelServer/Skills/Combat/MagnumShot.cs
+++ b/src/ChannelServer/Skills/Combat/MagnumShot.cs
@@ -201,7 +201,7 @@ namespace Aura.Channel.Skills.Combat
 			}
 
 			// Reduce arrows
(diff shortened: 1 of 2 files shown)


exectails:
Black Dragon drop? Stupid new content, making me fix stuff =<

>does not get removed on skill use

Uhm... isn't that the point?

>- Adds 15 minimum damage and 30 maximum damage.
- After normal damage calculations, another 15% damage is applied.
- Contains an **infinite amount of arrows**.
- Can only be used with the Bhafel Hunter.

> `originalName: "Bhafelsepaer Huntress", tags: "[...]/unlimited_arrow/`

exectails:
Hm... the proper fix for this is to check for the `/unlimited_arrow/` tag. If possible, you should never hard code item ids, even if currently only one item uses a function, that could change any time. I'll let you do this, for practice^^

xeroplz:
^ Fixed

PR #53, opened 2015-10-14
Updated Duncan

xCryptic:
• Added missing keywords
• Added missing shop info
• Other small fixes

### DIFF
--- a/system/scripts/npcs/tir/duncan.cs
+++ b/system/scripts/npcs/tir/duncan.cs
@@ -150,9 +150,13 @@ public class DuncanBaseScript : NpcScript
 				break;
 
 			case "about_arbeit":
-				Msg("Unimplemented");
+				Msg("Are you interested in a part-time job?<br/>It's great to see young people eager to work!<br/>To get one, talk to the people in town with the 'Part-Time Jobs' keyword.<br/>If you go at the right time, you'll be offered a job.");
+				Msg("If you do a good job, you will be duly rewarded.<br/>Just make sure to return to the person who gave you the job and report the results before the deadline.<br/>If you miss the deadline, you will not be rewarded regardless of how hard you worked.");
+				Msg("Part-time jobs aren't available 24 hours a day.<br/>You have to get there at the right time.");
+				Msg("The sign-up period usually begins between 7:00 am and 9:00 am.<br/>Since there are only a limited number of jobs available,<br/>others may take them all if you're too late.<br/>Also, you can do only one part-time job per day.");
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tir/duncan.cs
```diff
@@ -90,52 +101,217 @@ public override void Load()
...
 
 			case "about_arbeit":
-				Msg("I don't have any jobs for you, but you can get a part time job in town.");
-				//Msg("Are you interested in a part-time job?<br/>It's great to see young people eager to work!<br/>To get one, talk to the people in town with the 'Part-Time Jobs' keyword.<br/>If you go at the right time, you'll be offered a job.<p/>If you do a good job, you will be duly rewarded.<br/>Just make sure to return to the person who gave you the job and report the results before the deadline.<br/>If you miss the deadline, you will not be rewarded regardless of how hard you worked.<p/>Part-time jobs aren't available 24 hours a day.<br/>You have to get there at the right time.<p/>The sign-up period usually begins between 7:00 am and 9:00 am.<br/>Since there are only a limited number of jobs available,<br/>others may take them all if you're too late.<br/>Also, you can do only one part-time job per day.<p/>It looks like Nora and Caitin could use your help,<br/>so head to the Grocery Store or the Inn and talk to them.<br/>Start the conversation with them with the keyword 'Part-Time Jobs' and make sure it's between 7 and 9 am.<br/>Good luck!");
+				Msg("Unimplemented");
```
exectails: Why did you change it to "Unimplemented"? Does Duncan have a PTJ now? o.o If so, you should add a feature check.

xCryptic: I've been changing them all to unimplemented. I recall awhile back you told me they should say Unimplemented instead of "come back later" (for example https://github.com/aura-project/aura/commit/8396c6e9843a20865a98d599c77dbf503f4c5ab3#diff-b5b7895529c93fb6363184c4f27b2ed7R162)

exectails: Yea... if they actually have a PTJ, Duncan doesn't, why would it say "Unimplemented"? xD"

xCryptic: Uhh.. I don't get what you're saying. Duncan doesn't have one, so it *should* say Unimplemented. The only NPC right now that does have a ptj is Endelyon. So every other NPC will say Unimplemented.

exectails: ... Duncan does *not* offer a PTJ ***officially***, there is no PTJ that could be implemented for Duncan, so it should *not* say "Unimplemented", because there never will be one. That's why he explains that you should go to others to see if they got some, that's the *official* response.

PR #56, opened 2015-10-22
Added Event Props to the Campfire Skill

xCryptic:


src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -152,7 +155,7 @@ public void Complete(Creature creature, Skill skill, Packet packet)
 			// Set up Campfire
 			var pos = new Position(positionId);
 			var effect = (skill.Info.Rank < SkillRank.RB ? "campfire_01" : "campfire_02");
-			var prop = new Prop(PropId, creature.RegionId, pos.X, pos.Y, MabiMath.ByteToRadian(creature.Direction), 1); // Logs
+			var prop = new Prop(getPropId(creature), creature.RegionId, pos.X, pos.Y, MabiMath.ByteToRadian(creature.Direction), 1); // Logs
```
exectails: This won't work if you use up the the campfire kit (it's decremented before this line, and you're getting the item in the function). Also, passing a creature to a method that's supposed to give you a prop id, based on an item is really confusing.

Change the method to take an Item, create a variable `propId` at the beginning of this method, set it to the default prop id, and set it again to the result of the new `GetPropId` that you call inside the kit if/else, where you have access to the item, without asking for it again, and without risk of it not being there anymore.

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -197,5 +200,34 @@ private int GetDuration(SkillRank rank, int regionId)
 
 			return duration;
 		}
+
+		/// <summary>
+		/// Gets the prop ID
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <returns></returns>
+		public int getPropId(Creature creature)
```
exectails: This method should be private, since it's not needed anywhere else, and the "g" should be uppercase.

src/ChannelServer/Skills/Life/Campfire.cs
xCryptic: If I pass item through, what do I assign it as? I thought it would be a string but guess not :<

QewQew: Assign what? An item is an item. Just pass it into the function like any other item.

xCryptic: Oh, silly me.

xCryptic:
Ah crap, forgot to change the param

src/ChannelServer/Skills/Life/Campfire.cs
QewQew: That doesn't fix the larger issue at hand that exec mentioned.

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -197,5 +200,33 @@ private int GetDuration(SkillRank rank, int regionId)
...
 		}
+
+		/// <summary>
+		/// Gets the prop ID
+		/// </summary>
+		/// <param name="item"></param>
+		/// <returns></returns>
+		private int GetPropId(Item item)
+		{
+			var propId = PropId;
```
DeleteIT: IMO, it's pretty pointless to have this variable assigned, when the objective is simply to return one of four values.

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -197,5 +200,33 @@ private int GetDuration(SkillRank rank, int regionId)
...
+
+			if (item != null)
+			{
+				if (item.HasTag("/halloween_campfire_kit/"))
```
DeleteIT: Why not a switch, with a default case for prop 203?

DeleteIT: Because I'm forgetting that this is basically `Contains([kitname])`
Nubery already achieved for the day :+1: 

In any case, you can do `if (item.HasTag([tag])) return RelevantID`

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -127,6 +130,7 @@ public void Complete(Creature creature, Skill skill, Packet packet)
 			var positionId = packet.GetLong();
 			var unkInt1 = packet.GetInt();
 			var unkInt2 = packet.GetInt();
+			var item = creature.Inventory.GetItem(creature.Temp.CampfireKitItemEntityId);
```
exectails: While this doesn't produce an error, it's a pointless action. `CampfireKitItemEntityId` will be 0 if no kit was used, so you're searching the entire inventory for an item that can't exist.

Also, `CampfireKitItemEntityId` is currently not reset anywhere, someone could use a kit first, and firewood afterwards, but because you moved this here, it would use the kit's item again to determine the prop.

src/ChannelServer/Skills/Life/Campfire.cs
xCryptic: Good idea, dunno why I didn't think of a switch statement in the first place.

QewQew: How are you going to achieve this switch statement?

xCryptic: Well, I have an idea, not sure how well it'll work. Are you implying it's better to leave it as is?

DeleteIT: No, he's implying that it's effectively impossible to construct a switch statement out of boolean responses.

src/ChannelServer/Skills/Life/Campfire.cs
xCryptic: Hmmm.. well, this is what happens when I don't understand everything fully. What do you suggest I do?

DeleteIT:
```
Item item;

if (creature.Temp.CampfireKitItemEntityIdFuckThisIsSoDamnLong != 0)
{
item = creature.Inventory.GetItem(creature.Temp.CampfireKitItemEntityId);
creature.Temp.CampfireKitItemEntityIdFuckThisIsSoDamnLong = 0;
}
```

Would be the immediate solution.

DeleteIT: Additionally, doesn't
`creature.Temp.CampfireKitItemEntityId = MabiDictionary.Fetch<long>("ITEMID", dict);` let you use basically anything in the place of an actual campkit?

I've not looked at the packet, so I'm basing that on the fact that there's no checks.

exectails: I already told you the easiest way to do it o,o Why so complicated?

exectails: Hm, I though I wrote a comment earlier about the easiest way to do it, guess not. Just define a propId var at the beginning, set it to default, and set it again inside the kit else, where you have access to the item.

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -142,19 +148,20 @@ public void Complete(Creature creature, Skill skill, Packet packet)
 			{
 				// Check kit
 				var item = creature.Inventory.GetItem(creature.Temp.CampfireKitItemEntityId);
+
 				if (item == null)
 					throw new ModerateViolation("Used CampfireKit with invalid kit.");
 
+				prop = new Prop(GetPropId(item), creature.RegionId, pos.X, pos.Y, MabiMath.ByteToRadian(creature.Direction), 1); // Logs
```
exectails: Why are you creating two props now? There is no reason to do that work twice.

xCryptic: Oh. Why didn't you point this out before >:o

exectails: Because you didn't do this before, and I told you the cleanest possible way to do it =<

xCryptic: Yeah I did. The most recent update (today) only fixed the fire effect so it doesn't show on the Halloween campfire kit. :l

src/ChannelServer/Skills/Life/Campfire.cs
```diff
@@ -142,19 +148,20 @@ public void Complete(Creature creature, Skill skill, Packet packet)
 			{
 				// Check kit
 				var item = creature.Inventory.GetItem(creature.Temp.CampfireKitItemEntityId);
+
 				if (item == null)
 					throw new ModerateViolation("Used CampfireKit with invalid kit.");
 
+				prop.Data.Id = GetPropId(item); // Change the prop ID based on what campfire kit was used
```
exectails: Data is the entry of the prop in the prop db. What you're changing here would have erroring consequences. Never change anything in the Data property. Out of curiosity, why aren't you doing it the way I suggested 2 days or so ago?

xCryptic: Weird, I tested it and didn't get any errors. 

I don't even know anymore. You said to define the default prop before the if, then when there's access to the item, modify it - which is basically the structure I'm using.

I really don't know enough to be doing something like this. You might as well just use the function I made and implement the rest.

exectails: That's not what I said, what I did say was define a prop**Id** var with the default value. It's not about knowing stuff, it's about approaching the problem in the right way. Think about what you want to do, what you have, and what you need.

You need to create a prop, with a specific id, and you only want to do that once, because anything else would be redundant and inefficient. The only instance where you're not using the default id, is if a kit is used, in which case (inside the else) you have what you need to decide which prop id to use, the kit item. This means you'll need a variable in the scope of the prop declaration, that you can change when appropriate, which is then finally used to create the prop.

It is a simple problem once you wrap your head around it, you have a default value, that changes under certain circumstances, so define the default value, change it if appropriate, and then use it.

xCryptic: Ohhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh I get it now. Should I just trash this pile of mess and re-submit it?

exectails: Please do so^^

PR #64, opened 2015-11-04
Fixed "addskill" Magic Word

QewQew:


exectails:
Weird, I know I fixed this before... was that in a branch I didn't merge...? Hm. Anyway, thanks.

PR #67, opened 2015-11-05
Corrected stamina consumption for Healing skill

### LINKED ISSUE #66: Stamina consumption upon healing friendly target
Title says it all, stamina will be consumed when healing a friendly target. This is not how I remember it. However, stamina will be consumed if heal is used on your own character. Right now it is just consumed either way.

QewQew:
Stamina is not consumed if Healing is used on other players.
Fixes #66.


exectails:
Speedy fix, thanks^^

Pegwald:
good job :)

Issue #81, opened 2015-11-06
Dungeon Instance bug


Xcelled:
Bug video: https://youtu.be/wkJskScv9Mc

When a dungeon is created, party A is set as the "creators" of the dungeon. Dungeon creators *do not change* once the dungeon has been constructed.

Party "A" is then warped in to the dungeon. This currently works, but it works for the wrong reason. Instead of the *party* being warped in, this list of *creators* is. At this point, creators is the same as Party A, so there's no visible effect.

When party B comes along and tries to enter the dungeon, the mistake in the warping code becomes obvious. Once again, instead of warping the *party*, the *creators* are warped. However, creators is really the members of Party **A**, not Party **B**, so party B is left on the altar while A is warped back into the dungeon. You can clearly see this in the bug video.

In summary:
 - Altar check code is fine
 - Dungeon constructor/creation is fine, aside from a slightly confusing member name.
 - Warping into a dungeon is *incorrect* because *the wrong set of creatures are warped*.

Xcelled:
As I currently lack commit access...

http://puu.sh/lbaAu/b8b0d71acd.patch

QewQew:
Pull request. :+1:

Xcelled:
Would require forking and then moving my commits to the fork, which I'm too lazy to do... ;)

PR #83, opened 2015-11-06
Dgn instance fix

Xcelled:
Fixes that hilarious bug, aka #81.

### DIFF
--- a/src/ChannelServer/Util/GmCommands.cs
+++ b/src/ChannelServer/Util/GmCommands.cs
@@ -1614,7 +1614,7 @@ namespace Aura.Channel.Util
 				return CommandResult.InvalidArgument;
 			}
 
-			if (!ChannelServer.Instance.World.DungeonManager.CreateDungeonAndWarp(dungeonName, itemId, target))
+			if (!ChannelServer.Instance.World.DungeonManager.CreateDungeonAndWarp(dungeonName, itemId, target, null))
 			{
 				Send.SystemMessage(sender, Localization.Get("Failed to create dungeon."), dungeonName);
 				return CommandResult.Fail;
--- a/src/ChannelServer/World/Dungeons/DungeonManager.cs
+++ b/src/ChannelServer/World/Dungeons/DungeonManager.cs
@@ -286,7 +286,7 @@ namespace Aura.Channel.World.Dungeons
 		/// <param name="dungeonName"></param>
 		/// <param name="itemId"></param>
 		/// <param name="creature"></param>
-		/// <param name="clientEvent"></param>
+		/// <param name="altarEvent"></param>
 		/// <returns></returns>
 		public bool CreateDungeonAndWarp(string dungeonName, int itemId, Creature creature, ClientEvent altarEvent)
 		{
@@ -309,7 +309,7 @@ namespace Aura.Channel.World.Dungeons
 						}
 					}
 
-					// Original creature is always a member of the dungeon party.
+					// Don't use Dungeon.Party as it doesn't work with instances.
 					foreach (var member in toWarp)
 					{
 						var pos = member.GetPosition();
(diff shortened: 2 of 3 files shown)


exectails:
This seems hackish. There's code in place that was supposed to select the players inside the event already, that code should be fixed instead.

Xcelled:
I didn't see any in the immediate vicinity. I'll look again, but the original code (obviously) didn't check properly either.

exectails:
>that code should be fixed instead.

...

Xcelled:
I'm saying "that code" may not exist xD

exectails:
I am looking at it... I wouldn't say "there is code" if there wasn't.

Xcelled:
Link?

exectails:
https://github.com/aura-project/aura/blob/master/src/ChannelServer/World/Dungeons/Dungeon.cs#L159

Xcelled:
That's for something different, though Dungeon.Party should really be named Dungeon.Creators, if I understand the code. That code sets the people who initially entered the dungeon. I don't think anything needs to be fixed there, it's working properly, though we might want to change it to a HashSet.

This issue is with the warp code into an existing dungeon. Original code warped all the dungeon *creators* into the dungeon, instead of warping the party that dropped the item.

I could change the CreateAndWarp code to accept a creature enumerable to control who gets warped, and then filter the party in the caller method. This would remove the null in the command, and provide a more intuitive interface.

exectails:
>This issue is with the warp code into an existing dungeon. Original code warped all the dungeon creators into the dungeon, instead of warping the party that dropped the item.

The "creators" are the members of the party of the creature who dropped the item that were standing on the altar. And that's precisely the group of people who should be warped into the dungeon.

Xcelled:
You're not hearing me xD. CREATING a new dungeon works fine. ENTERING an existing dungeon does not, because it *incorrectly warps the creators again*.

Creating works by "coincidence". When you create a dungeon, `toWarp == creators`, obviously. But when entering an existing dungeon, `toWarp = partyThatJustDroppedAnItem`. The original code was flawed because it always used creators, not PartyThatJustDroppedAnItem.

exectails:
Well, you didn't do a good job explaining what you fixed for what reasons.

>Dgn instance fix
>Fixes that hilarious bug.

So... why get any new information at all? You have the creature, you have its party, the party has a method to get all members who're on the altar, excluding the leader (which maybe should be changed), why not reuse that?

Xcelled:
When a dungeon is created, party A is set as the "creators" of the dungeon. Dungeon creators *do not change* once the dungeon has been constructed. This is the code you linked to, and is working fine.

Party "A" is then warped in to the dungeon. This currently works, but it works for the wrong reason. Instead of the *party* being warped in, this list of *creators* is. At this point, creators is the same as Party A, so there's no visible effect.

When party B comes along and tries to enter the dungeon, the mistake in the warping code becomes obvious. Once again, instead of warping the *party*, the *creators* are warped. However, creators is really the members of Party **A**, not Party **B**, so party B is left on the altar while A is warped back into the dungeon. You can clearly see this in the bug video I posted on #81 .

In summary:
 - Altar check code is fine
 - Dungeon constructor/creation is fine, aside from a slightly confusing member name.
 - Warping into a dungeon is *incorrect* because *the wrong set of creatures are warped*.

exectails:
Thanks for the explanation, though I did get it by now^^ What about my question?

Xcelled:
Because I didn't see that method, and then I fell into a headfirst slide down the dungeon code.

exectails:
I've fixed this now. You didn't fix the `Send.EntitiesDisappear` hack btw. Almost forgot that as well. (There are certainly some things that need to get cleaned up/fix/unhack in Aura.)

PR #84, opened 2015-11-06
Dgn exists msg

Xcelled:
Adds the dungeon instance already exists warning.

### DIFF
--- a/src/ChannelServer/Util/GmCommands.cs
+++ b/src/ChannelServer/Util/GmCommands.cs
@@ -1614,7 +1614,7 @@ namespace Aura.Channel.Util
 				return CommandResult.InvalidArgument;
 			}
 
-			if (!ChannelServer.Instance.World.DungeonManager.CreateDungeonAndWarp(dungeonName, itemId, target))
+			if (!ChannelServer.Instance.World.DungeonManager.CreateDungeonAndWarp(dungeonName, itemId, target, null))
 			{
 				Send.SystemMessage(sender, Localization.Get("Failed to create dungeon."), dungeonName);
 				return CommandResult.Fail;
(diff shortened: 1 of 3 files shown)


exectails:
I wonder if it might be cleaner to add a member to Dungeon, maybe a usage counter, so you can tell whether it's used for the first time or not without adding an out parameter.

exectails:
Or maybe... check the actual list of creators, that would make a lot of sense. If you go into a dungeon instance again, that you created, do you get the msg as well?

Xcelled:
A player count wouldn't be enough. Someone could have just left the dungeon and you get their instance again, with no warning.

I already only send the message to non-creators: https://github.com/aura-project/aura/blob/9217892849bd64d3a0843d814bf93e792b773fb0/src/ChannelServer/World/Dungeons/DungeonManager.cs#L329

You can't use the creator list as a reliable creation check either. While it works in this case, you're still losing the essential information of "was a dungeon created?". In short, I don't see a problem with the out, as it's a clean solution.

exectails:
>A player count wouldn't be enough.

I mean a counter how many times the dungeon was used o.o

>You can't use the creator list as a reliable creation check either. While it works in this case, you're still losing the essential information of "was a dungeon created?".

Do you have an example for why you would need that information?

Xcelled:
>I meant a counter of how many times the dungeon was used o.o

Do you have an example for why you would need that information? ^^

I don't see either approach having a compelling advantage over the other, and it's so small we can always change it later if it becomes a problem.

exectails:
You're basically doing the wrong thing here, that's what I'm trying to say. You're checking for whether the dungeon is new, and if it is, you filter the creatures that should get the msg. That *works*, but it's not the deciding factor. Now it's there, it doesn't really matter, but it shouldn't have been added in the first place, that was unnecessary work and made the code less intuitive.

This includes commits that don't even belong in this PR. Will you make a new one, so I can merge it?

exectails:
Fixed in 400a44a974e04de2c867529780cdb476cb16a3e4.

PR #86, opened 2015-11-07
AI reset aggro if target is warping

### LINKED ISSUE #85: Insane knockback bug
Meanwhile in Tir, near the moongate by Ferghus, I was testing the knockdown skill requirements for the magic bolt spells when I found myself in an itchy situation. I was down to like 10 hp and decided to run away to try to escape my inevitable death of a white wolf that was running at me with smash. Ahead of me was the entrance to Dugald Aisle, which would be my great escape... or that's what I was expecting.

This is very hard for me to explain since I did catch it on tape this time, but I will give it a shot. I am not joking with you when I am about to tell this.
Upon reaching the entrance, I clicked it and thought I was safe. However, the wolf managed to catch up to me and landed its smash while my screen was fading to black. When it faded back in, I could not recognize the place around me. It wasn't before I opened my map I figured what had happened.

https://youtu.be/CJjCzWsVFGM
This is the only footage I got of this incident.

Do we take this seriously or? I am just speechless.

QewQew:
Fixes #85?

### DIFF
--- a/src/ChannelServer/Scripting/Scripts/AiScript.cs
+++ b/src/ChannelServer/Scripting/Scripts/AiScript.cs
@@ -351,6 +351,7 @@ namespace Aura.Channel.Scripting.Scripts
 			// Reset on...
 			if (this.Creature.Target.IsDead																 // target dead
 			|| !this.Creature.GetPosition().InRange(this.Creature.Target.GetPosition(), _aggroMaxRadius) // out of aggro range
+			|| this.Creature.Target.Warping																 // target is warping
 			|| this.Creature.Target.Client.State == ClientState.Dead									 // target disconnected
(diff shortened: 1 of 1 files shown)


exectails:
Halfway there, you should also check if the target is warping in the target selection, so it doesn't even get selected as target. Otherwise it *would* select that target and then switch back on the next tick.

exectails:
Hm, what I mean were these:
```
var hated = potentialTargets.Where(cr => !cr.IsDead && this.DoesHate(cr) && !cr.Has(CreatureStates.NamedNpc));
```
But your solution is better, there's no reason to even consider them potential targets. We could actually move more things into the `Where` you added, like `IsDead` :+1:

PR #87, opened 2015-11-08
Added system message text wrapping.

Xcelled:
Mabi can only show 100 chars per message. System messages, like command
help (esp the output of go) can exceed 100 chars and are just cut off.

This commit adds a word-wrapping algorithm and changes system messages so
the message is first wrapped by 100 chars (Mabi's limit) and then sent in
multiple messages.

Before:

After:

### DIFF
--- a/src/ChannelServer/Network/Sending/Send.Messages.cs
+++ b/src/ChannelServer/Network/Sending/Send.Messages.cs
@@ -41,15 +41,21 @@ namespace Aura.Channel.Network.Sending
 		public static void System_Broadcast(string from, string format, params object[] args)
 		{
 			var packet = new Packet(Op.Chat, MabiId.Broadcast);
-			packet.PutByte(0);
-			packet.PutString("<{0}>", from);
-			packet.PutString(format, args);
-			packet.PutByte(true);
-			packet.PutUInt(0xFFFF8080);
-			packet.PutInt(0);
-			packet.PutByte(0);
 
-			ChannelServer.Instance.World.Broadcast(packet);
+			foreach (var msg in string.Format(format, args).Chunkify(100)) // Mabi displays up to 100 chars
+			{
+				packet.PutByte(0);
+				packet.PutString("<{0}>", from);
+				packet.PutString(msg);
+				packet.PutByte(true);
+				packet.PutUInt(0xFFFF8080);
+				packet.PutInt(0);
+				packet.PutByte(0);
+
+				ChannelServer.Instance.World.Broadcast(packet);
+
+				packet.Clear(packet.Op, packet.Id);
+			}
 		}
 
 		/// <summary>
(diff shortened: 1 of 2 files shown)


Xcelled:
Question: Should this be added to player chat as well? On the one hand, it would be nice, cause some NPC phrases exceed 100 chars. On the other, a pake-er could easily send a 100k char string and cause a DoS if not an outright crash.

Actually do we even truncate chat length?? A 100k string could still cause a crash even now, as the memory buffers overload. This could happen on receive too, not just send...

QewQew:
Why not just split NPC phrases only and leave player chat alone?

Xcelled:
That's one option, yes, but it would have to be done in NPC script.

I confirmed the presence of the chat length bug, see #88.

exectails:
>Actually do we even truncate chat length??

We don't atm, we should definitely add a check for it.

>A 100k string could still cause a crash even now

I don't know what the client will do with a 4k message, but that's the worst that can happen. The receive buffer is 4096 long, you can't possibly send a 100k long packet to the server. Or rather, it wouldn't be able to handle it, log an exception, and kill the client.

I see the appeal for system messages, but imo chat shouldn't be run through this. There is an official limit, that plays into the look&feel, I think we should stick with that. It would be weird if you say something, it's split in 2, and you only see the second half above your character.

QewQew:
Probably the upcoming issue.

Xcelled:
Decided not to make an issue after all. Basically we're missing a lot of bounds checks all over the place. These could cause issues with crafted packets and crashing, esp on low memory systems. A couple examples from the packet class, because that's where I happen to be:

- GetBin reads a short from the packet, does 0 range checks, and then allocates a byte[] of that many.
- GetObj is worse for a different, but related, issue. If the PtrToStructure call throws an exception (because the struct is not of type T or someone crafted a packet) the allocated unmanaged memory is leaked.
- ReadVarInt is vulnerabile to overflow. Not sure how much this really impacts tho.
- Likewise no range checks on GetString()

It's possible to starve Aura, 4kb at a time :)

Xcelled:
>I see the appeal for system messages, but imo chat shouldn't be run through this. There is an official limit, that plays into the look&feel, I think we should stick with that. It would be weird if you say something, it's split in 2, and you only see the second half above your character.

I think the client displays up to 300 chars in the chat bubble.

I agree about the players. They couldn't take advantage of it anyway without mods/hacks, but NPCs we might want to consider. There are at least two NPCs I know of on NA who got their text cut off (the dude in the elf arena and that angry guy in the Scat base camp), so it's an issue, esp with localization.

exectails:
>so it's an issue, esp with localization.

Be that as it may, splitting it doesn't solve the issue^^" Instead of cutting off the latter half, you'd be cutting off the first half. Or you'd have to split it and then send it a few seconds apart.

>GetObj

GetObj isn't used btw, at least not in the server. It only exists because it's needed in Pale and clients.

Xcelled:
>Be that as it may, splitting it doesn't solve the issue^^" Instead of cutting off the latter half, you'd be cutting off the first half. Or you'd have to split it and then send it a few seconds apart.

No, if you send three large chats, you get a large chat over your head o.o

QewQew:
>Maybe we should add support for \n in phrases, so people have additional control over where the split happens? (ie, they can split it early, after a period, for a more natural flow)

That seems like a feature that'll never be used.

exectails:
>No, if you send three large chats quickly, you get a large chat over your head o.o

Oh~ right, I forgot about that.

>That seems like a feature that'll never be used.

Indeed.

Splitting those long NPC phrases you were talking about would probably make sense. We should add a length check in the Chunk function though, before actually splitting and looping, I'd feel bad about doing that for every message, even though very few will ever need it.

PR #88, opened 2015-11-09
Updated Endelyon

xCryptic:
• Added Missing Keywords
• Added Upgrading
• Updated Keywords
• Other small fixes

### DIFF
--- a/system/scripts/npcs/tir/endelyon.cs
+++ b/system/scripts/npcs/tir/endelyon.cs
@@ -97,7 +97,7 @@ public class EndelyonScript : NpcScript
 				break;
 
 			case "about_skill":
-				if (!QuestActive(204002))
+				if (!QuestActive(204002) && !QuestCompleted(204002))
 				{
 					//StartQuest(204002); // Quest log: http://pastebin.com/1c9XFc5Y
 					Msg("Skills? I don't know if I can call it a skill, but you can gather eggs from hens.<br/>Some of the less fortunate among the faithful come to the Church seeking the blessings of Lymilark.<br/>To relieve their hunger, we prepare boiled eggs for them.");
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tir/endelyon.cs
```diff
@@ -67,19 +86,35 @@ public override void Load()
...
 				GiveKeyword("shop_healing");
-				Msg("<face name='normal'/>Have you met Dilys? She's the town's Healer.<br/>Walk along the road heading northeast, and you will find the Healer's House.<br/>Make sure to meet her if you pass by there.<br/>");
+				Msg("<face name='normal'/>Have you met Dilys? She's the town's Healer.<br/>Walk along the road heading northeast, and you will find the Healer's House.<br/>Make sure to meet her if you pass by there.");
 				ModifyRelation(Random(2), 0, Random(2));
 				break;
 
 			case "about_skill":
-				GiveKeyword("skill_tailoring");
-				Msg("I'm sorry, but I don't know much about skills.<br/>The only skill I know is tailoring.<br/>If you want to know more about making clothes,<br/>talk with Caitin at the Grocery Store.<br/>She's the best tailor in town.<br/>");
+				if (!QuestActive(204002))
```
exectails: You probably got this from the official scripts, but it seems wrong. It goes to the Tailoring only while this quest is active? What if you were to finish this quest and not talk to her, you would never get to the Tailoring part. Something isn't right here. Could you check the code again? Maybe it checks for whether the player finished this quest yet?

xCryptic: Hmmm, you're right.


I sorta missed that first part since I thought it was just localization stuff. I should be checking if it has been completed and if it's in progress.

exectails:
Alright. Naturally, the quest dialog will be moved to the quest scripts when they get added, but this is fine for now. It's good to have the dialog ready.

Issue #143, opened 2015-12-20
Refining training r6


Pegwald:
Ranking this skill is a pain without having access to mythril, but the real pain here is that I am not able to get any training from refining iron. I am supposed to get 0.05 exp per success, but I don't get anything at all.

Pegwald:
It works normally ^^
GJ

PR #144, opened 2015-12-20
Fixed Refining Rank 6 skill traning

QewQew:
#143


exectails:
If you write `Fixed #144`, or `Close #144`, in either summary or description, it will automatically close the issue when the PR is merged.

PR #152, opened 2015-12-24
Added Tarlach + Bear

Sairii:
Missing any quest related stuff or Ego:<


system/scripts/npcs/sidhe/tarlach_bear.cs
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		End("You have ended your conversation with <npcname/>.");
+	}
+
+	protected override async Task Keywords(string keyword)
+	{
+		switch (keyword)
+		{
+			case "personal_info":
```
exectails: Did you put the same random messages in both on purpose? If so, why?

exectails:
You should always keep in mind that every single commit will appear in the commit list. The following is *very* undescriptive, if you look at the second and third commit separately. Always write the summary as if you only did that one commit, summarizing what you did in that one commit.

Sairii:
oh well no idea why I did that

PR #155, opened 2015-12-24
fixed #147

### LINKED ISSUE #147: Dungeon boss-room generation bug
It has happened twice now that whenever I am about to do a boss of a dungeon, I am greeted with an already open boss door, but no boss. I was not able to find any key and the treasure room is still locked. If it's necessary, I do have some footage of it that I can share with you.

Tachiorz:


exectails:
This will make sure that there's a open boss door, right? But isn't it more of a hot-fix? I mean, the actual problem isn't solved, but when you now get to the door it will simply open/you'll be warped in?

Tachiorz:
Yeah, this makes just an unlocked boss door. There was a rare case with issue #147 when God of random made rolls of ores and there wasn't any space left for puzzle with lock
I was thinking about rerolling floor or section few times until boss door is locked. But cleaning up after puzzles is such a pain -_-

exectails:
It should be a pretty rare problem... still kinda unfortunate =/ Well, better a hot-fix than no fix for now.

PR #156, opened 2015-12-24
Updated Malcom

xCryptic:
• Added missing keywords
• Added greetings
• Added repairing
• Other small code fixes/updates

Github has a weird way of showing changes in the shop o.o

### DIFF
--- a/system/scripts/npcs/tir/malcolm.cs
+++ b/system/scripts/npcs/tir/malcolm.cs
@@ -4,7 +4,7 @@
 // General store manager
 //---------------------------------------------------------------------------
 
-public class MalcolmBaseScript : NpcScript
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tir/malcolm.cs
```diff
@@ -170,92 +328,112 @@ public override void Load()
 	}
 }
 
+//Add("General Goods", 1006, "MGCSEL:s:yellow;MGCWRD:s:knowskill(10005);"); // Introduction to Music Composition
```
exectails: ?

xCryptic: Just leaving that there in case we still need the metadata.

PR #158, opened 2015-12-25
Gairech fieldspawns

Sairii:
Added Gairech fieldspawns. 
Before exec is doing more stuff that I already have.

The wisp AI is missing the stacks of lightningbolt.
I dont know if the AIs can handle that now. I only remember that it couldnt back then.

Also im done for today... i have to wake up fairly early tmrw...
I'll add more tho. So dont do things I already did xD


exectails:
Why are some coordinates rather different from officials and why are the delays missing for several spawns?

Sairii:
o.O wat

Sairii:
I added the new gairech file with corrected delays.
It also still has the group IDs for easier overview.

PR #162, opened 2015-12-26
Rock throwing

UnknownVoid:
Working Rock Throwing skill, No other changes.

Might require some improvment but should be good.


src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+namespace Aura.Channel.Skills.Action
+{
+	/// <summary>
+	/// Rock Throwing handler
+	/// </summary>
+	[Skill(SkillId.RockThrowing)]
+	public class RockThrowing : ISkillHandler, IPreparable, IReadyable, ICompletable, ICancelable, ICombatSkill
+	{
+		private const int KnockBackDistance = 100; //might be greater then this from killing the npcs with rock throwing they do fly back slightly..
+		private const int TargetStun = 500;
```
exectails: Why have two variables if they have the same value?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+		private const int TargetStunElf = 500;
+		private const int AttackerStun = 1500;
+		private const float StabilityReduction = 0f; // only specified due to packets /
```
exectails: If stability isn't reduced, why even have the code for it?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+	public class RockThrowing : ISkillHandler, IPreparable, IReadyable, ICompletable, ICancelable, ICombatSkill
+	{
+		private const int KnockBackDistance = 100; //might be greater then this from killing the npcs with rock throwing they do fly back slightly..
+		private const int TargetStun = 500;
+		private const int TargetStunElf = 500;
+		private const int AttackerStun = 1500;
+		private const float StabilityReduction = 0f; // only specified due to packets / 
+		private const float StabilityReductionElf = 0f;
+		
+        /// <summary>
```
exectails: Spaces.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+		public CombatSkillResult Use(Creature attacker, Skill skill, long targetEntityId)
+		{
+			var target = attacker.Region.GetCreature(targetEntityId);
+			if (target == null)
+				return CombatSkillResult.InvalidTarget;
+
+			var targetPos = target.GetPosition();
+			var attackerPos = attacker.GetPosition();
+
+			var actionType = (attacker.IsElf ? CombatActionPackType.ChainRangeAttack : CombatActionPackType.NormalAttack);
```
exectails: The action type is actually different between elves and humans/other races...?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+			var attackerStun = (short)(actionType == CombatActionPackType.ChainRangeAttack ? TargetStunElf : AttackerStun);
+
+			Send.SkillUse(attacker, skill.Info.Id, attackerStun, 1);
+
+			var prevId = 0;
```
exectails: ...? I assume you c&p'ed the CombatMastery or RangedAttack handler and didn't know what this is? It's needed for dual wielding, where the following code is potentially executed multiple times, in a loop, you can safely remove it.

Oh... yea, you definitely copied RangedAttack. Hm o~o I'm afraid you didn't do a good job removing the RangedAttack specific code.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+				cap.Type = actionType;
+				cap.PrevId = prevId;
+				prevId = cap.Id;
+
+				var aAction = new AttackerAction(CombatActionType.RangeHit, attacker, targetEntityId);
+				aAction.Set(AttackerOptions.Result);
+				aAction.Stun = attackerStun;
+				cap.Add(aAction);
+
+				var targetSkillId = target.Skills.ActiveSkill != null ? target.Skills.ActiveSkill.Info.Id : SkillId.CombatMastery;
```
exectails: The skill id you pass to TargetAction is of the attacker's skill, not the target's. The target's skill id is set automatically.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+
+				var tAction = new TargetAction(CombatActionType.TakeHit, target, attacker, targetSkillId);
+				tAction.Set(TargetOptions.Result);
+				tAction.AttackerSkillId = skill.Info.Id;
```
exectails: You don't have to set this, as you are, or should be, passing the information via the constructor.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+				tAction.Stun = (short)(actionType == CombatActionPackType.NormalAttack ? TargetStunElf : TargetStun);
+				if (actionType == CombatActionPackType.NormalAttack)
+					tAction.EffectFlags = 0x20;
+				cap.Add(tAction);
+
+				// Damage
+				var damage = target.GetRndRangedDamage();
```
exectails: Rock Throwing does default range damage? Doesn't the Wiki say it doesn't do any damage?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+
+				// Critical Hit
+				var critChance = attacker.GetRightCritChance(target.Protection);
```
exectails: Do you even need to handle Crits, ManaShield, etc, for a skill that supposedly doesn't do damage?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+						if (target.IsUnstable)
+						{
+							tAction.Set(TargetOptions.KnockBack);
+							attacker.Shove(target, KnockBackDistance);
+						}
+					}
+					tAction.Creature.Stun = tAction.Stun;
+				}
+
+				aAction.Creature.Stun = aAction.Stun;
```
exectails: This is done automatically in the CombatActionPack handling.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,199 @@
...
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Complete(Creature creature, Skill skill, Packet packet)
+		{
+			Send.SkillComplete(creature, skill.Info.Id);
+		}
+
+		/// <summary>
+		/// Cancels skill, Cancels Motion and returns the character to idle position.
```
exectails: That's not what this method does.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,163 @@
...
+	/// <summary>
+	/// Rock Throwing handler
+	/// </summary>
+	[Skill(SkillId.RockThrowing)]
+	public class RockThrowing : ISkillHandler, IPreparable, IReadyable, ICompletable, ICancelable, ICombatSkill
+	{
+		private const int KnockBackDistance = 100; //might be greater then this from killing the npcs with rock throwing they do fly back slightly..
+		private const int TargetStun = 500;
+		private const int AttackerStun = 1500;
+        private const int StabilityReduction = 0;
```
exectails: Spaces.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,163 @@
...
+        public CombatSkillResult Use(Creature attacker, Skill skill, long targetEntityId)
+        {
+            var target = attacker.Region.GetCreature(targetEntityId);
+            if (target == null)
+                return CombatSkillResult.InvalidTarget;
+
+            var targetPos = target.GetPosition();
+            var attackerPos = attacker.GetPosition();
+
+            var actionType = (attacker.IsElf ? CombatActionPackType.ChainRangeAttack : CombatActionPackType.NormalAttack);
```
exectails: You're still differentiating between different races and action types, which is most likely not a thing in Rock Throwing.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,163 @@
...
+                // Actions
+                var cap = new CombatActionPack(attacker, skill.Info.Id);
+                cap.Type = actionType;
+
+                var aAction = new AttackerAction(CombatActionType.RangeHit, attacker, targetEntityId);
+                aAction.Set(AttackerOptions.Result);
+                aAction.Stun = attackerStun;
+                cap.Add(aAction);
+
+                var targetSkillId = target.Skills.ActiveSkill != null ? target.Skills.ActiveSkill.Info.Id : SkillId.CombatMastery;
```
exectails: Still redundant, as it's handled by the TargetAction ctor.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,163 @@
...
+
+                var tAction = new TargetAction(CombatActionType.TakeHit, target, attacker, targetSkillId);
+                tAction.Set(TargetOptions.Result);
+                tAction.AttackerSkillId = skill.Info.Id;
+                tAction.Stun = (short)(actionType == CombatActionPackType.ChainRangeAttack ? AttackerStun : TargetStun);
+                if (actionType == CombatActionPackType.ChainRangeAttack)
```
exectails: Is this actually used for Rock Throwing?

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// Prepares skill.
+        /// </summary>
+        /// <param name="creature"></param>
+        /// <param name="skill"></param>
+        /// <param name="packet"></param>
+        /// <returns></returns>
+        public bool Prepare(Creature creature, Skill skill, Packet packet)
+        {
+            creature.StopMove();
+            creature.Lock(Locks.Walk | Locks.Run);
```
exectails: The skill handlers handle this.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,140 @@
...
+                cap.Add(tAction);
+
+                // TODO: Add Ability attack Puzzle Orbs / Activate them
+                // No Idea how to do that..
+
+                // Aggros
+                target.Aggro(attacker);
+
+                // Death/Knockback
+                if (target.IsDead)
```
exectails: Why check for death if you're not doing any damage?

UnknownVoid: Mostly only for when you change the finish rules, then it's possible to finish the enemy with rock throwing when it's dead and waiting to be finished.

Though, I will remove this.

src/ChannelServer/Skills/Actions/RockThrowing.cs
UnknownVoid: without enforcing this lock, I was able to bypass the skills animations. Well,  I will remove it.

exectails:
>>` +            creature.Lock(Locks.Walk | Locks.Run);`

>without enforcing this lock, I was able to bypass the skills animations. Well, I will remove it.

If that's the case, we have to figure out *why* that is, because the server is supposed to basically do the same exact same thing before calling Prepare.

```js
{ id: 50020, name: "Rock Throwing", masterTitle: 0, type: 0, prepare: { lock: "Walk|Run" }
```

>> `+                if (target.IsDead)`

>Mostly only for when you change the finish rules, then it's possible to finish the enemy with rock throwing when it's dead and waiting to be finished.

>Though, I will remove this.

Perfectly valid reason to leave it in there, though it should have a TODO or something then, so *others* know what it does there.

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,148 @@
...
+        /// <param name="skill"></param>
+        /// <param name="packet"></param>
+        /// <returns></returns>
+        public bool Prepare(Creature creature, Skill skill, Packet packet)
+        {
+            creature.StopMove();
+
+            //TODO: Test animation bug
+            // without enforced Locks for some skills the client allows the player to bypass the animations that are currently playing (such as grabing the rock / knock down)
+            creature.Lock(Locks.Walk | Locks.Run);
```
exectails: Any idea why this would be required, if basically the exact same thing happens a ms before, in the SkillPrepare handler? =/ Like, literally 4 actual lines before this?

src/ChannelServer/Skills/Actions/RockThrowing.cs
exectails: .

src/ChannelServer/Skills/Actions/RockThrowing.cs
```diff
@@ -0,0 +1,148 @@
...
+        {
+            var target = attacker.Region.GetCreature(targetEntityId);
+            if (target == null)
+                return CombatSkillResult.InvalidTarget;
+
+            var targetPos = target.GetPosition();
+            var attackerPos = attacker.GetPosition();
+
+            var actionType = (attacker.IsElf ? CombatActionPackType.ChainRangeAttack : CombatActionPackType.NormalAttack);
+            var attackerStun = (short)(actionType == CombatActionPackType.ChainRangeAttack ? AttackerStun : AttackerStun);
```
exectails: No need for the ternary if if you use the same value both times^^

exectails:
Any news on this PR? It's been sitting here a while, and as far as I can see you've still got ranged code in there. And what about that orb TODO, is there a problem with attacking orbs?

UnknownVoid:
The only issues remaining in this skill is that it can't attack dungeon orbs like in official to "activate" the orbs, Code wise I'm not sure how to accomplish this and I'm unsure if there is even any methods currently for skills in aura to do that.

I will close this PR and keep this skill private.

PR #173, opened 2015-12-28
Added direwolf.cs AI

Sairii:
also added the AI in the races.txt


system/scripts/ai/direwolf.cs
```diff
@@ -0,0 +1,124 @@
+//--- Aura Script -----------------------------------------------------------
+//  Direwolf AI
+//--- Description -----------------------------------------------------------
+//  AI for direwolf creatures.
+//--- Missing ---------------------------------------------------------------
+//  Wolf- Support is missing.
```
exectails: What does that mean? Keep in mind that others have to understand your comments.

PR #180, opened 2015-12-30
Added Ciar Beginner Golem AI & DB entry

Sairii:
The Do(Wait(500)); is custom. It does make the AI run smoother visually.


Sairii:
All the goblin AIs are minimal different.... it makes me wanna puke.
Also I forgot to mention that I changed the Knockcount of GoldGoblin from 2 to 3.
That was the only way to use Attack(). GoldGoblin should attack 3 times with the fluted sword. And the same AI is used for Poison Goblins which only hit twice.
Which is why I made a Timeout Attack().

exectails:
>Also I forgot to mention that I changed the Knockcount of GoldGoblin from 2 to 3 in the DB.
That was the only way to use Attack(). GoldGoblin should attack 3 times with the fluted sword.

`Attack(3)`? o,o The KnockCount in the race db doesn't even affect anything if the creature has a weapon.

system/scripts/ai/ciarbeginnergolem.cs
```diff
@@ -0,0 +1,115 @@
+//--- Aura Script -----------------------------------------------------------
+// Ciar Beginner Golem AI
+//--- Description -----------------------------------------------------------
+// AI for Ciar Beginner Golem (ID 130014)
+//---------------------------------------------------------------------------
+
+[AiScript("ciarbeginnergolem")]
```
exectails: Didn't we agree on "smallgolem"?

system/scripts/ai/goblin.cs
```diff
@@ -0,0 +1,63 @@
...
+[AiScript("goblin")]
+public class GoblinAi : AiScript
+{
+	public GoblinAi()
+	{
+		SetAggroRadius(850); // angle 120 audio 200
+		Hates("/pc/", "/pet/");
+		Hates("/ahchemy_golem/");
+		Hates("/rp/");
+		SetAggroLimit(AggroLimit.None);
```
exectails: Really? *No limit'? That means in a dungeon room every single goblin would aggro you o.o

system/scripts/ai/ciarbeginnergolem.cs
Sairii: did we?

system/scripts/ai/goblin.cs
Sairii: Official AI says party... wiki says unkown...
tho spider fighter types which also use this AI all have single except one hardmode one has multi.
You are right tho I've never seen them ganking up on you.

system/scripts/ai/ciarbeginnergolem.cs
Sairii: there is more than one small golem. which is why this one was named small golem 3.

exectails: That's precisely what I said =<

system/scripts/ai/goblin.cs
exectails: We've discussed `party` in the past, but iirc we weren't able to figure what it means. In this case it probably shouldn't be multi aggro, especially not unlimited.

Sairii: trying to remember what else had party

Sairii: direwolf has party , grizzly has party, gremlin has party, wisp has party

system/scripts/ai/goblin.cs
```diff
@@ -0,0 +1,63 @@
...
+	protected override IEnumerable Aggro()
+	{
+		Do(KeepDistance(400, false, 2000));
+		Do(Circle(300, 700, 1000, false));
+		var rndAggro = Random();
+		if (rndAggro < 60) // 60%
+		{
+			Do(SwitchTo(WeaponSet.First));
+			Do(Attack(2, 5000));
+			Do(Wait(500)); // Custom slow down
```
exectails: There's nothing "custom" about it, if this is how it feels right in *our* AI. This doesn't change what the AI is supposed to do, it's simply the code require to make it do that.

Sairii:
So the question is.... what party aggro means

system/scripts/ai/goblin.cs
Sairii: I added that custom comment because maybe something gets changed in the future that changes the pace of attack?
Dunno.

exectails:
I'd rather say the question is what their behavior is. If a Goblin is single aggro, that's what we should do.

Sairii:
Yes but other party types are Multi

exectails:
That was an example.

Sairii:
Yeah but I still have to figure out others then...
We do have grizzly bear with none same with wisp but I doubt thats right...
So imma go and play with grizzlys and look for the Black Wizard fieldboss... caus i need loads of wisps

Sairii:
I just hope there arent any more party types.... because the wiki aint helping with "multi" slapped on each mob...

exectails:
Well, I'd expect all party mobs to behave pretty much the same. Why would two AIs do different things with the same target type?

PR #194, opened 2016-01-06
Kobold AIs

Sairii:
Archer coming too in the next minutes.


Sairii:
Uhm something is wrong with the monsters ranged attack damage.
I already noticed it while doing goblin archers.
The damage is not official, it is really low.
10-60 damage is a kobold archer supposed to do. But it crits for 23 and most of the time hits for only 9. It is really puny dmg.

system/scripts/ai/kobold.cs
```diff
@@ -0,0 +1,66 @@
...
+		Do(Wander(100, 500));
+		Do(Wait(2000, 5000));
+	}
+
+	protected override IEnumerable Aggro()
+	{
+		Do(KeepDistance(400, true, 2000));
+		Do(Circle(300, 800, 1000, false));
+
+		SwitchRandom();
```
exectails: Hah~ I love it.

system/scripts/ai/koboldarcher.cs
```diff
@@ -0,0 +1,76 @@
...
+		else if (Case(40))
+		{
+			Do(PrepareSkill(SkillId.MagnumShot));
+			Do(RangedAttack());
+		}
+	}
+
+	private IEnumerable OnKnockDown()
+	{
+		if (Case(30))
```
exectails: This won't work without `SwitchRandom()`, or rather, the result will be kind of random, because the Cases will use the last value generated by SwitchRandom the last time.

Sairii:
wait... WAT

Sairii:
omfg i must be too tired for this

exectails:
:+1:

system/scripts/ai/koboldarcher.cs
Sairii: Hmm wierd that I didnt get any errors and the AI worked without problem.
I didnt trigger knockdown tho so maybe only that didnt work, who knows

PR #208, opened 2016-01-13
Shutdownv2

superjeffreyc:
Main fixes from the previous pull request:

- Fixed formatting
- New internal op for shutting down
- Verify that the shutdown command comes from the LoginServer or from a player with admin rights

Below is what was listed in the previous pull request:

The shutdown command has the following usage: shutdown [seconds]
- The minimum time is 60 seconds and the maximum time is 1800 seconds.
- An internal broadcast is sent to notify all players and a MsgBox is sent to each player.
- The shutdown command on the Login console will shutdown all channels, while the GM shutdown command will only shutdown that channel.
- All user data will be saved shortly before kicking all users out.
- The channel will refuse and disconnect all incoming connections once the shutdown command is issued.
- Prevented multiple shutdown commands for a channel.
- The ChannelServer will close once the shutdown is complete, putting the channel into "Maintenance".
- LoginServer, WebServer, and MsgrServer are not closed in this process.


src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+		{
+			if (client.Address.Equals(ChannelServer.Instance.Conf.Channel.LoginHost + ":" + ChannelServer.Instance.Conf.Channel.LoginPort))
+			{
+				Log.Info("Address confirmed to be login server. Proceeding with shutdown.");
+				Shutdown(time);
+			}
+			else
+			{
+				Log.Error("Address does not match the login server. Rejecting shutdown request.");
+				return;
```
QewQew: redundant return statement

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			// Shutdown warning
+			Send.Internal_Broadcast(Localization.Get("The server will be brought down for maintenance in " + time + " seconds. Please log out safely before then."));
+
+			// Save a list of all clients logged in prior to shutdown
+			ChannelServer.Instance.ShutdownClientList = ChannelServer.Instance.Server.Clients;
+
+			// Send MsgBox to all users
+			foreach (var user in ChannelServer.Instance.ShutdownClientList)
+			{
+				try
```
QewQew: Would a lock here be more appropriate? What exceptions would you get?

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+
+			// Get stored list of all clients logged in prior to shutdown and kill
+			this.KillConnectedClients();
+
+			// Double check that nobody is left (in case someone logged in while the server was switching to shutdown mode)
+			ChannelServer.Instance.ShutdownClientList = ChannelServer.Instance.Server.Clients;
+			this.KillConnectedClients();
+
+			Log.Info("Waiting an additional 10 seconds to save data before closing...");
+			Thread.Sleep(10 * 1000);
```
QewQew: Isn't everything running on this thread?

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			}
+
+			Log.Info("Shutting down in {0} seconds...", time);
+
+			timer = new Timer(new TimerCallback(ShutdownTimerDone));
+			timer.Change(time * 1000, Timeout.Infinite);
+
+		}
+
+		private void ShutdownTimerDone(object timer)
```
QewQew: `timer` hides previously declared `timer`.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+				catch (Exception e)
+				{
+					Log.Exception(e, "Failed to send MsgBox to user.");
+				}
+			}
+
+			Log.Info("Shutting down in {0} seconds...", time);
+
+			timer = new Timer(new TimerCallback(ShutdownTimerDone));
+			timer.Change(time * 1000, Timeout.Infinite);
```
QewQew: Why not just put this in the constructor?

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+		}
+
+		private void ShutdownTimerDone(object timer)
+		{
+			((Timer)timer).Dispose();
+
+			// Get stored list of all clients logged in prior to shutdown and kill
+			this.KillConnectedClients();
+
+			// Double check that nobody is left (in case someone logged in while the server was switching to shutdown mode)
```
QewQew: You should have disabled logins by this time.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			Log.Info("Waiting an additional 10 seconds to save data before closing...");
+			Thread.Sleep(10 * 1000);
+			CliUtil.Exit(0, false);
+		}
+
+		private void KillConnectedClients()
+		{
+			foreach (var user in ChannelServer.Instance.ShutdownClientList)
+			{
+				try
```
QewQew: Locks?

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -1888,6 +1889,30 @@ private CommandResult HandlePoints(ChannelClient client, Creature sender, Creatu
...
+			if (!int.TryParse(args[1], out time))
+				return CommandResult.InvalidArgument;
+
+			if (ChannelServer.Instance.ShuttingDown)
+			{
+				Send.MsgBox(client.Controlling, Localization.Get("Server is already being shut down."));
+				return CommandResult.Fail;
+			}
+
+			time = Math2.Clamp(60, 1800, time);
```
QewQew: If you're going to clamp to a min of 60, you might as well make the `time` parameter optional.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
 				// TODO: clean up dbs
 			}
 		}
+
+		public void Shutdown(int time, ChannelClient client)
+		{
+			if (client.Address.Equals(ChannelServer.Instance.Conf.Channel.LoginHost + ":" + ChannelServer.Instance.Conf.Channel.LoginPort))
```
QewQew: For some reason, this just doesn't stick well with me.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			}
+		}
+
+		public void Shutdown(int time)
+		{
+
+			this.ShuttingDown = true;
+
+			// Shutdown warning
+			Send.Internal_Broadcast(Localization.Get("The server will be brought down for maintenance in " + time + " seconds. Please log out safely before then."));
```
QewQew: Use formatting strings.

src/ChannelServer/Network/ChannelClient.cs
```diff
@@ -113,6 +113,9 @@ public override void CleanUp()
 			foreach (var creature in this.Creatures.Values)
 				creature.Dispose();
 
+			// Prevent additional packets from being sent to this client since it is being removed shortly
+			this.State = ClientState.Dead;
```
QewQew: CleanUp() is used in DisconnectRequest, and a response packet is sent immediately after cleaning up.
This prevents the response from ever reaching the client.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -1888,6 +1889,30 @@ private CommandResult HandlePoints(ChannelClient client, Creature sender, Creatu
...
+			if (args.Count < 2)
+				return CommandResult.InvalidArgument;
+
+			// Get time
+			if (!int.TryParse(args[1], out time))
+				return CommandResult.InvalidArgument;
+
+			if (ChannelServer.Instance.ShuttingDown)
+			{
+				Send.MsgBox(client.Controlling, Localization.Get("Server is already being shut down."));
```
QewQew: sender

src/ChannelServer/ChannelServer.cs
ImaBrokeDude: Wouldn't this just freeze the server for 10 seconds? Nothing would happen during this time.

QewQew: This is on a timer thread.

ImaBrokeDude:
If I may comment, don't forget about your ability to use 

```
// This example will safely log the client out in 30 seconds with an automatic notification view MsgBox
hannelServer.Instance.World.Broadcast(new Packet(Op.RequestClientDisconnect, MabiId.Channel).PutInt(30 * 1000));
```

QewQew:
Except, probably not used exactly like that.

src/ChannelServer/ChannelServer.cs
ImaBrokeDude: Gotcha thanks, just kinda..... skipped that detail there hehe.

src/ChannelServer/ChannelServer.cs
```diff
@@ -70,6 +71,9 @@ public class ChannelServer : ServerMain
 		public PartyManager PartyManager { get; private set; }
 
 		public WorldManager World { get; private set; }
+		public bool ShuttingDown { get; private set; }
+		private List<ChannelClient> ShutdownClientList { get; set; }
+		private Timer timer { get; set; }
```
exectails: By convention, private variables have to prefixed with an underscore, while properties should be pascal case and are *usually* used for publics in Aura. Both are fine, but you should stick to the naming conventions^^

src/ChannelServer/ChannelServer.cs
exectails: Me neither... I also don't particularly like how it's done inside here, it feels like it should go into the handler.

A better solution for confirming the source might be to simply check if the client is the same as ChannelServer.LoginServer. Alternatively, you could use the inter password for authorization, this is the kind of thing it was added for.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+				Shutdown(time);
+			}
+			else
+			{
+				Log.Error("Address does not match the login server. Rejecting shutdown request.");
+				return;
+			}
+		}
+
+		public void Shutdown(int time)
```
exectails: You should document what "time" is.

src/ChannelServer/ChannelServer.cs
exectails: ^ Because otherwise this wouldn't be localizable. `Localization.Get` looks up that *exact* string you're passing it.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+		{
+
+			this.ShuttingDown = true;
+
+			// Shutdown warning
+			Send.Internal_Broadcast(Localization.Get("The server will be brought down for maintenance in " + time + " seconds. Please log out safely before then."));
+
+			// Save a list of all clients logged in prior to shutdown
+			ChannelServer.Instance.ShutdownClientList = ChannelServer.Instance.Server.Clients;
```
exectails: You're not saving a list here, you're saving the reference to the list, which might still change at any moment. I assume you wanted a copy to make sure the list doesn't change half-way through, but you have to call `ToList` on it, or something similar.

src/ChannelServer/ChannelServer.cs
exectails: A lock? What do you want lock here?

Since he's iterating over the raw clients, there are a bunch of exceptions you might get. For example, he's accessing user.Controlling, but that might not be set yet, he's sending to creature.Client in MsgBox, but the client might be in the process of signing off. Though I think it would be better to explicitly check for these things, and maybe not output some of these exceptions. You could easily flood the log, for no real reason.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			timer.Change(time * 1000, Timeout.Infinite);
+
+		}
+
+		private void ShutdownTimerDone(object timer)
+		{
+			((Timer)timer).Dispose();
+
+			// Get stored list of all clients logged in prior to shutdown and kill
+			this.KillConnectedClients();
```
exectails: Why are you "saving" a list at the start of the timer, to kill all clients here? They're informed beforehand so they can safely log out, which many will probably do. You don't want to kill the clients that were online back then, but now. So there's actually no reason for the global variable ShutdownClientList. KillConnectedClients could actually get it itself.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +298,82 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+
+			// Double check that nobody is left (in case someone logged in while the server was switching to shutdown mode)
+			ChannelServer.Instance.ShutdownClientList = ChannelServer.Instance.Server.Clients;
+			this.KillConnectedClients();
+
+			Log.Info("Waiting an additional 10 seconds to save data before closing...");
```
exectails: White there's nothing wrong with waiting a moment, I wanted to mention that technically it's not needed. You're killing all client in a loop, for every client CleanUp is called, which saves all data. This a synchronous process, and there's nothing left to wait for once you're out of the loop.

Speaking of saving things, you didn't save the global scripting vars anywhere, did you? Those are normally saved every 5 minutes, and have to be saved after the players have been logged out, otherwise things like seal stones might get rolled back, if something happened in the last 5 minutes. See OnMabiTick in the ScriptManager.

src/LoginServer/Network/Send.cs
```diff
@@ -620,5 +620,16 @@ private static void Add(this Packet packet, Account account)
 
 			packet.PutByte(0);
 		}
+
+		/// <summary>
+		/// Sends shutdown request to all channels.
+		/// </summary>
+		/// <param name="time">Seconds until shutdown.</param>
+		public static void ShutdownRequest(int time)
```
exectails: By convention, the sender method is usually named just like the Op.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -1888,6 +1889,30 @@ private CommandResult HandlePoints(ChannelClient client, Creature sender, Creatu
...
+				return CommandResult.Fail;
+			}
+
+			time = Math2.Clamp(60, 1800, time);
+
+			ChannelServer.Instance.Shutdown(time);
+
```
exectails: You should tell the one executing the command that something is going to happen now.

superjeffreyc:
Thanks again for the feedback. I'll be working on these changes now.

src/ChannelServer/Util/GmCommands.cs
superjeffreyc: Everyone will be notified about the automatic logout through a MsgBox sent by the ChannelServer so I don't think it is necessary here.

src/ChannelServer/ChannelServer.cs
```diff
@@ -70,6 +73,9 @@ public class ChannelServer : ServerMain
 		public PartyManager PartyManager { get; private set; }
 
 		public WorldManager World { get; private set; }
+		public bool ShuttingDown { get; private set; }
+		private List<ChannelClient> _shutdownclientlist { get; set; }
```
exectails: Private *fields* get an underscore prefix, *properties*, no matter their visibility, get PascalCase.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +302,55 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+
+			this.KillConnectedClients();
+
+			CliUtil.Exit(0, false);
+		}
+
+		private void KillConnectedClients()
+		{
+			// Grab a copy of the list of users still currently logged in
+			ChannelServer.Instance._shutdownclientlist = ChannelServer.Instance.Server.Clients.ToList<ChannelClient>();
```
exectails: You are inside the instance of ChannelServer, there's really no reason for the `ChannelServer.Instance.`.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +302,55 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+
+			// Kill all clients still logged in
+			foreach (var user in ChannelServer.Instance._shutdownclientlist)
```
exectails: This method is the only place you use this list, why even make a global copy?

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +302,55 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+						user.Kill();
+				}
+				catch (Exception e)
+				{
+					Log.Exception(e, "Error killing client.");
+				}
+			}
+
+			// Save global variables
+			this.ScriptManager.OnMabiTick(ErinnTime.Now);
```
exectails: Hm =/ While this works now, if we were to suddenly decide to do something odd in side that method, this call from here might give weird results. You should only do what you want to do.

src/ChannelServer/Network/Sending/Send.Messages.cs
```diff
@@ -187,6 +187,15 @@ public static void PartyAdChat(Party party)
 		}
 
 		/// <summary>
+		/// Sends MsgBox to creature's client notifying automatic logout.
+		/// </summary>
+		/// <param name="time">Time in seconds until automatic logout.</param>
+		public static void MsgBox(int time)
```
exectails: You shouldn't use the name of a completely different op, for the method that sends RequestClientDisconnect. When reading the code that uses the method, you want to be able to tell what is sent. I was really confused just now, when I saw a call to MsgBox, which normally doesn't take an int parameter.

src/ChannelServer/Util/GmCommands.cs
exectails: Oh, you're sending that packet immediately... yea, I guess that's okay then. In the version I wrote years ago I waited with the force dc until 5 minutes or so before the shutdown. I think officials do something similar.

src/LoginServer/Util/LoginConsoleCommands.cs
```diff
@@ -12,14 +13,33 @@ public class LoginConsoleCommands : ConsoleCommands
 	{
 		public LoginConsoleCommands()
 		{
-			this.Add("shutdown", "Orders all servers to shut down", HandleShutDown);
+			this.Add("shutdown", "<seconds>", "Orders all servers to shut down", HandleShutDown);
```
exectails: If the parameter is optional it should be in square brackets.

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +301,55 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+			this.ShuttingDown = true;
+
+			Send.Internal_Broadcast(Localization.Get(String.Format("The server will be brought down for maintenance in {0} seconds. Please log out safely before then.", time)));
+
+			Send.RequestClientDisconnect(time);
+
+			Log.Info("Shutting down in {0} seconds...", time);
+
+			this.Timer.Change(time * 1000, Timeout.Infinite);
+
```
exectails: Shouldn't you put the channels in maintenance or something when the shutdown command comes in, so no people can connect anymore? They wouldn't get the broadcast nor the disconnect request.

src/ChannelServer/Network/Sending/Send.Messages.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Aura development team - Licensed under GNU GPL
+﻿// Copyight (c) Aura development team - Licensed under GNU GPL
```
exectails: o.o?

src/ChannelServer/ChannelServer.cs
```diff
@@ -294,5 +301,68 @@ public override void InitDatabase(AuraDb db, BaseConf conf)
...
+				Log.Warning("Unregistered channel.");
+			}
+			else
+			{
+				channel.State = ChannelState.Maintenance;
+				Send.Internal_ChannelStatus_Maintenance();
+				Log.Info("{0} switched to maintenance.", this.Conf.Channel.ChannelName);
+			}
+
+			Send.Internal_Broadcast(Localization.Get(String.Format("The server will be brought down for maintenance in {0} seconds. Please log out safely before then.", time)));
```
exectails: You have to wrap the *string* in the localization call, not the format. It would try to look up the exact string, with the seconds if you did it like this.

src/ChannelServer/Network/Sending/Send.Internal.cs
```diff
@@ -41,6 +41,26 @@ public static void Internal_ChannelStatus()
 		}
 
 		/// <summary>
+		/// Sends Internal.ChannelStatus to login server notifying change of state to maintenance.
+		/// </summary>
+		public static void Internal_ChannelStatus_Maintenance()
```
exectails: This is redundant, why not make `Internal_ChannelStatus` take a parameter with the status?

exectails:
Looking good overall =)

exectails:
Thanks for your contribution.

PR #236, opened 2016-02-16
Added Lightning Rod Skill

xeroplz:
- Adds lightning rod skill.
- I'm really not sure about that workaround in Network/Handlers/Skills.cs though, it's so bad but the only way I can think of at the moment... However the workaround behaves exactly like Lightning Rod does on official, making me believe devCAT did exactly the same thing.


src/ChannelServer/Network/Handlers/Skills.cs
```diff
@@ -255,7 +255,9 @@ public void SkillPrepare(ChannelClient client, Packet packet)
 				// TODO: Use regens, for shiny gradually depleting bars
 				if (skill.RankData.ManaCost != 0 || skill.RankData.StaminaCost != 0)
 				{
-					creature.Mana -= skill.RankData.ManaCost;
+					if (skill.Info.Id != SkillId.LightningRod) // Lightning Rod is excluded
```
exectails: Maybe you could generalize this more, not hardcoding a specific skill id... though I'm not sure what that would look like. What exactly is it that makes Lightning Rod special? And how could you check for that? Obviously we're doing something wrong/different from officials here, otherwise we wouldn't have a problem, so in the long run we have to change how we do it for a proper implementation, but figuring out the difference is the first step.

src/ChannelServer/Network/Handlers/Skills.cs
```diff
@@ -382,6 +384,14 @@ public void SkillUse(ChannelClient client, Packet packet)
 				return;
 			}
 
+			// Lightning Rod Use Check
```
exectails: This comment is lacking a little^^ What *exactly* does this check do? How can ActiveSkill be != skill at this point? And what does canceling do?

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+	[Skill(SkillId.LightningRod)]
+	public class LightningRod : ISkillHandler, IPreparable, IUseable, ICompletable, ICancelable, IInitiableSkillHandler
+	{
+		/// <summary>
+		/// Length of attack area; unconfirmed.
+		/// </summary>
+		/// <remarks>
+		/// "http://wiki.mabinogiworld.com/view/Lightning_Rod#Summary"
+		/// </remarks>
+		private const int skillLength = 1400;
```
exectails: According to our coding conventions these consts should be PascalCased.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+		/// <summary>
+		/// Uses LightningRod
+		/// </summary>
+		/// <param name="attacker"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Use(Creature attacker, Skill skill, Packet packet)
+		{
+			// Mana degeneration
+			var degen = attacker.Regens.Add(Stat.Mana, (-1 * skill.RankData.ManaCost), attacker.ManaMax, degenTime);
```
exectails: You might want to use ManaActive instead of negating ManaCost, that's the value that is supposed to be used for this. Was the change in the packet handler just to get the animated mana reduction?

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+			var p2 = new Point(pivot.X - skillLength / 2, pivot.Y + skillWidth / 2);
+			var p3 = new Point(pivot.X + skillLength / 2, pivot.Y + skillWidth / 2);
+			var p4 = new Point(pivot.X + skillLength / 2, pivot.Y - skillWidth / 2);
+			p1 = this.RotatePoint(p1, pivot, r);
+			p2 = this.RotatePoint(p2, pivot, r);
+			p3 = this.RotatePoint(p3, pivot, r);
+			p4 = this.RotatePoint(p4, pivot, r);
+
+			// TargetProp
+			var LProp = new Prop(280, attacker.RegionId, poe.X, poe.Y, MabiMath.ByteToRadian(attacker.Direction), 1f, 0f, "single");
```
exectails: According to our coding conventions this normal variable should be camelCased.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+
+			var targetAreaId = new Location(attacker.RegionId, poe).ToLocationId();
+
+			var aAction = new AttackerAction(CombatActionType.SpecialHit, attacker, targetAreaId);
+			aAction.Set(AttackerOptions.KnockBackHit1 | AttackerOptions.UseEffect);
+			aAction.PropId = LProp.EntityId;
+			cap.Add(aAction);
+
+			// Get targets in Polygon
+			var targets = attacker.Region.GetCreaturesInPolygon(p1, p2, p3, p4).Where(x => attacker.CanTarget(x)).ToList();
```
exectails: Is CanTarget enough here? No collision checks for example?

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+					{
+						tAction.Set(TargetOptions.KnockDown);
+						attacker.Shove(target, knockbackDistance);
+					}
+				}
+				tAction.Creature.Stun = tAction.Stun;
+			}
+			cap.Handle();
+
+			Send.Effect(attacker, Effect.LightningRod, 3, poe.X, poe.Y); // Lightning Shooting Effect
```
exectails: You might want to make an enum for the third parameter, for better readability and less comments.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,367 @@
...
+			switch (attackerSkill.Info.Rank)
+			{
+				case SkillRank.RF:
+				case SkillRank.RE:
+					attackerSkill.Train(2); // Attack an enemy
+					break;
+				case SkillRank.RD:
+				case SkillRank.RC:
+					attackerSkill.Train(2); // Attack an enemy
+					if (action.Attacker.Temp.LightningRodFullCharge == true) attackerSkill.Train(3); // Attack an Enemy with a Max Charge
```
exectails: The value already is a bool, no need for `== true`.

src/ChannelServer/Network/Handlers/Skills.cs
xeroplz: When you use lightning rod and you don't have enough mana, the skill get's canceled. However, when you release the button, the client sends use regarless of whether the skill was canceled or not. When that happens, "oh no! you've tried to use a skill that wasn't active!" *client lock*. I had to manually set ActiveSkill in order for proper cancel handling to take place.

src/ChannelServer/Network/Handlers/Skills.cs
xeroplz: Lightning Rod uses its mana on Use, not prepare.

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Done :heavy_check_mark:

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Done :ballot_box_with_check: :

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Done :ballot_box_with_check:

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Done :ballot_box_with_check:

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Done :ballot_box_with_check:

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Huh... I thought that was automatic o-o
I don't recall ever hitting something through a wall or anything, but I'll need to check again.

src/ChannelServer/Skills/Magic/LightningRod.cs
QewQew: Instead of spamming people with worthless "Done"s
image
Don't.
We'll know by seeing if it's an outdated diff.

xeroplz: They've been removed :+1:

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,388 @@
...
+				ManaShield.Handle(target, ref damage, tAction);
+
+				// Apply Damage
+				target.TakeDamage(tAction.Damage = damage, attacker);
+
+				// Stun Time
+				tAction.Stun = TargetStun;
+
+				// Reduce stun, based on ping
+				if (delayReduction > 0)
```
exectails: If it always knocks down, like Firebolt, should the stun be reduced? That could make them start running mid-fly with MD3.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,388 @@
...
+				else
+				{
+					// Always knock down
+					if (target.Is(RaceStands.KnockDownable))
+					{
+						tAction.Set(TargetOptions.KnockDown);
+						attacker.Shove(target, KnockbackDistance);
+					}
+				}
+				tAction.Creature.Stun = tAction.Stun;
```
exectails: The combat handler does this automatically.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,388 @@
...
+				case SkillRank.RC:
+					attackerSkill.Train(2); // Attack an enemy
+					if (action.Attacker.Temp.LightningRodFullCharge) attackerSkill.Train(3); // Attack an Enemy with a Max Charge
+					break;
+				case SkillRank.RB:
+				case SkillRank.RA:
+				case SkillRank.R9:
+				case SkillRank.R8:
+				case SkillRank.R7:
+                case SkillRank.R6:
```
exectails: Tabs.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,388 @@
...
+namespace Aura.Channel.Skills.Magic
+{
+	/// <summary>
+	/// Lightning Rod Handler
+	/// </summary>
+	/// Var1: Min Damage
+	/// Var2: Max Damage
+	/// Var3: Max Chargning Time (ms)
+	/// Var4: Max Charge Damage Bonus (%)
+	/// Var5: ?
```
exectails: Var5 is the cool down.

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: I'm unsure of what you're talking about (github won't show an exact line :<). What happens automatically?

src/ChannelServer/Skills/Magic/LightningRod.cs
xeroplz: Does Firebolt not get stun reduced on Mana Deflector? I always thought it did...

src/ChannelServer/Skills/Magic/LightningRod.cs
exectails: Line 242, setting the stun.

exectails:
About MD stun: the way I implemented it, MD doesn't reduce knock down stun, because enemies would stand up while they're being knocked back.

src/ChannelServer/Skills/Magic/LightningRod.cs
QewQew: 238 if you're looking at the newest version, btw.
And Github comments are always right below the line of mention, so... yeah.

src/ChannelServer/Skills/Magic/LightningRod.cs
```diff
@@ -0,0 +1,383 @@
...
+		/// <summary>
+		/// Prepares the skill
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		/// <returns></returns>
+		public bool Prepare(Creature creature, Skill skill, Packet packet)
+		{
+			if (creature.RightHand == null)
```
exectails: Didn't notice this before... is this right? You need *something* in your right hand? Not like it needs to be a staff or something (tag check)? And if it's not x, you send cancel, but continue through Prepare? Are you sure you shouldn't return false, so Prepare stops and sends the silent cancel itself?

PR #238, opened 2016-02-23
Improves ChannelStatus reporting

Seeker1437:
* ChannelStatus is calculated based off the comments for each
ChannelStatus enum.

* IsInMaintenance can be used to set the server to Maintenance easily since it
is something one might do in a script, the other states should happen
automatically.

// Questions

Should we hold the ChannelStatus as a property and allow changes that was with automatic updates on change? I think that would break the design of the PR but maybe it's better?

### DIFF
--- a/src/ChannelServer/ChannelServer.cs
+++ b/src/ChannelServer/ChannelServer.cs
@@ -20,6 +20,7 @@ using Aura.Shared.Util;
 using Aura.Shared.Util.Configuration;
 using System;
 using System.Net.Sockets;
+using System.Runtime.CompilerServices;
 using System.Threading;
 
 namespace Aura.Channel
@@ -226,6 +227,46 @@ namespace Aura.Channel
 		{
 			this.IsInMaintenance = isInMaintenance;
 			Send.Internal_ChannelStatus();
+		}
+
+		/// <summary>
+		/// Calculates the state of the channel.
+		/// </summary>
+		/// <remarks>
+		/// When calculating the <see cref="ChannelState"/> we take into account
+		/// whether the server is running as well as if it is in Maintenance.
+		/// </remarks>
+		/// <returns></returns>
+		public ChannelState CalculateChannelState()
+		{
+			if (this.IsInMaintenance)
+				// In case we do support the booting channel state
+				return this.IsRunning ? ChannelState.Maintenance : ChannelState.Booting;
+
+			double stress;
+			var current = this.World.CountPlayers();
+			var max = this.Conf.Channel.MaxUsers;
+
+			try
+			{
+				stress = (current / max) * 100;
+			}
+			catch (DivideByZeroException)
+			{
+				Log.Warning("Max user count was zero, falling back to Normal.");
+
+				// Fallback value
+				return ChannelState.Normal;
+			}
+
+			if (stress >= 40 && stress <= 70)
+				return ChannelState.Busy;
+			if (stress > 70 && stress <= 95)
+				return ChannelState.Full;
+			if (stress > 95)
+				return ChannelState.Bursting;
+
+			return ChannelState.Normal;
 		}
 
 		private void OnClientDisconnected(ChannelClient client)
--- a/src/ChannelServer/ChannelServer.cs
+++ b/src/ChannelServer/ChannelServer.cs
@@ -36,10 +36,7 @@ namespace Aura.Channel
 		/// </summary>
 		private const int LoginTryTime = 10 * 1000;
 
-		/// <summary>
-		/// Used to determine if the server is running
-		/// </summary>
-		public bool IsRunning { get; private set; }
+		private bool _isRunning;
 
 		/// <summary>
 		/// Gets or sets whether the server is in maintenance. 
@@ -112,7 +109,7 @@ namespace Aura.Channel
 		/// </summary>
 		public void Run()
 		{
-			if (this.IsRunning)
+			if (_isRunning)
 				throw new Exception("Server is already running.");
 
 			CliUtil.WriteHeader("Channel Server", ConsoleColor.DarkGreen);
@@ -156,7 +153,7 @@ namespace Aura.Channel
 			this.StartStatusUpdateTimer();
 
 			CliUtil.RunningTitle();
-			this.IsRunning = true;
+			_isRunning = true;
 
 			// Commands
 			this.ConsoleCommands.Wait();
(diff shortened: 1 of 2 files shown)


src/ChannelServer/ChannelServer.cs
```diff
@@ -33,7 +33,28 @@ public class ChannelServer : ServerMain
 		/// </summary>
 		private const int LoginTryTime = 10 * 1000;
 
-		private bool _running = false;
+		/// <summary>
+		/// Used to determine if the server is running
+		/// </summary>
+		public bool IsRunning { get; private set; } = false;
```
exectails: That's C# 6.

src/ChannelServer/ChannelServer.cs
```diff
@@ -33,7 +33,28 @@ public class ChannelServer : ServerMain
...
+
+		private bool _isInMaintenance = false;
```
exectails: `false` is the default value of a bool field.

src/ChannelServer/ChannelServer.cs
```diff
@@ -33,7 +33,28 @@ public class ChannelServer : ServerMain
...
+ 
+		/// <summary>
+		/// Can be used to set the server into maintenance
+		/// </summary>
+		public bool IsInMaintenance
```
exectails: Avoid sending packets from properties, especially if you don't document it.

src/ChannelServer/Network/Sending/Send.Internal.cs
```diff
@@ -35,12 +36,52 @@ public static void Internal_ChannelStatus()
...
+		/// Calculates the state of the channel.
+		/// </summary>
+		/// <remarks>
+		/// When calculating the <see cref="ChannelState"/> we take into account
+		/// whether the server is running as well as if it is in Maintenance.
+		/// </remarks>
+		/// <param name="current"></param>
+		/// <param name="max"></param>
+		/// <returns></returns>
+		private static ChannelState GetServerState(int current, int max)
```
exectails: Don't make it static and then access the instance you called from... one or the other. And don't forget the `this.` if you change it.

src/ChannelServer/Network/Sending/Send.Internal.cs
```diff
@@ -35,12 +36,52 @@ public static void Internal_ChannelStatus()
...
+
+			double stress;
+
+			try
+			{
+				stress = (current / max) * 100;
+			}
+			catch (Exception ex)
+			{
+				Log.Exception(ex, "Error occured when calculating server stress, falling back to Normal.");
```
exectails: There's a *single* thing that could wrong in that line, catch *that*, with a *proper* error message, not random exceptions with meaningless errors.

src/ChannelServer/Network/Sending/Send.Internal.cs
Seeker1437: Best recommended practice?

1. Extension Method
2. Move method inside ChannelServer
3. Add ChannelServer as parameter
4. Something I didn't think of?

src/ChannelServer/Network/Sending/Send.Internal.cs
```diff
@@ -35,12 +36,55 @@ public static void Internal_ChannelStatus()
...
+
+			double stress;
+
+			try
+			{
+				stress = (current / max) * 100;
+			}
+			catch (DivideByZeroException ex)
+			{
+				Log.Exception(ex, "Max user count was zero, falling back to Normal.");
```
exectails: Why log an entire exception if you know exactly what happened where?

src/ChannelServer/Network/Sending/Send.Internal.cs
```diff
@@ -35,12 +36,55 @@ public static void Internal_ChannelStatus()
...
+		/// <remarks>
+		/// When calculating the <see cref="ChannelState"/> we take into account
+		/// whether the server is running as well as if it is in Maintenance.
+		/// </remarks>
+		/// <param name="current"></param>
+		/// <param name="max"></param>
+		/// <param name="isRunning"></param>
+		/// <param name="isInMaintenance"></param>
+		/// <returns></returns>
+		private static ChannelState GetServerState(int current, int max, bool isRunning, bool isInMaintenance)
```
exectails: This doesn't belong in Send.

src/ChannelServer/ChannelServer.cs
```diff
@@ -210,6 +219,56 @@ public void ConnectToLogin(bool firstTime)
...
+
+		/// <summary>
+		/// Calculates the state of the channel.
+		/// </summary>
+		/// <remarks>
+		/// When calculating the <see cref="ChannelState"/> we take into account
+		/// whether the server is running as well as if it is in Maintenance.
+		/// </remarks>
+		/// <returns></returns>
+		public ChannelState GetServerState()
```
Seeker1437: Torn on this decision, but where I moved it to, makes it usable anywhere if it is at all used.

exectails:
I'll wait with merging this until the shutdown PR is done.

exectails:
Please adjust this PR for the recent changes, due to the merging of the shutdown command.

src/ChannelServer/ChannelServer.cs
```diff
@@ -36,7 +36,15 @@ public class ChannelServer : ServerMain
 		/// </summary>
 		private const int LoginTryTime = 10 * 1000;
 
-		private bool _running = false;
+		/// <summary>
+		/// Used to determine if the server is running
+		/// </summary>
+		public bool IsRunning { get; private set; }
```
exectails: Why did you make this public? You never even change it.

Seeker1437: This was the result of having CalculateChannelState() in the wrong place initially. I will correct this.

src/ChannelServer/ChannelServer.cs
```diff
@@ -36,7 +36,12 @@ public class ChannelServer : ServerMain
 		/// </summary>
 		private const int LoginTryTime = 10 * 1000;
 
-		private bool _running = false;
+		private bool _isRunning;
+
+		/// <summary>
+		/// Gets or sets whether the server is in maintenance. 
+		/// </summary>
+		public bool IsInMaintenance { get; set; }
```
exectails: Now that I think about it, what's even the point of this property?

src/ChannelServer/ChannelServer.cs
```diff
@@ -217,6 +222,60 @@ public void ConnectToLogin(bool firstTime)
...
+		/// <returns></returns>
+		public ChannelState CalculateChannelState()
+		{
+			// Just in case this gets called
+			if (this.ShuttingDown)               
+				return ChannelState.Maintenance;
+
+			if (this.IsInMaintenance)
+				// In case we do support the booting channel state
+				return this._isRunning ? ChannelState.Maintenance : ChannelState.Booting;
```
exectails: No this, and renaming the variable is a pointless change.

src/ChannelServer/ChannelServer.cs
```diff
@@ -217,6 +222,60 @@ public void ConnectToLogin(bool firstTime)
...
+
+			double stress;
+			var current = this.World.CountPlayers();
+			var max = this.Conf.Channel.MaxUsers;
+
+			try
```
exectails: There's no reason for this try catch, you know *exactly* what might go wrong, handle that, instead of just running the throwing calculations.

Issue #109, opened 2015-11-23
Command List Command


xCryptic:
I just noticed we don't have a command that shows a list of commands and what they do. I know you can just look at the config file, but what about the players on random servers? They won't know where to look.

So there should be a `>help` or `>commands` to show a list of commands (what they do and how to use them) based on the user's auth level.

Cosrnos:
This does sort of link back to #77 since the problem is how many commands are in the game. It wouldn't be a bad idea to "Organize" commands by type so you can do something like 

    > commands

    Available commands by type: `item` `player` `stat` `movement`

    > commands item

    drop - Drops an item at your current position
    item - places an item in your inventory
    info - finds item info

etc.

Xcelled:
I think #77 has broadened now - I'll retitle it and let's move all discussions there.

exectails:
Certainly not a bad suggestion, especially with normal players in mind, who won't get to see *all* the commands, but only the ones enabled for their authority level. Those players will probably never get a long list.

GMs/devs on the other hand would get a huge list, but I wonder if that's even a problem. You could categorize them, but really, they should have a list of commands anyway imo, without having to look them up with `help`. And for usage instructions all you need to do is type only the command, it will then fail and show you how to use it.

As Cryptic said, such a command would be most useful for normal players, on a random server.

ImaBrokeDude:
For something like this, would it do better to make the command type something that is set when the command is added so that the command itself will be more maintainable?

Something like

```
public enum CommandCategory : byte
{
	Item,
	Player,
	Stat,
	Movement
}

Add(50, 50, "item", "<id|name> [amount|color1 [color2 [color 3]]]", CommandCategory.Item, HandleWhere);
```

To accomplish:

```
> commands

Available commands by type: `item` `player` `stat` `movement`

> commands item

drop - Drops an item at your current position
item - places an item in your inventory
info - finds item info
```

DeleteIT:
`>help`
Check what commands the user has the privileges for, list all commands that meet criteria.
Do so to trade/whisper chat, with brief description + available aliases.

Alternatively, just list the basic commands they have access to, in trade/whisper chat, give a system message to prompt them to look at the trade/whisper channel, and then just allow users to do `>help [command name]` for the basic outline of what it does, and what aliases it has.

Preferably group them by the level those commands would be, user/gm/admin.


Anything else requires too much effort, and will ultimately prove confusing/time consuming to update, as very few commands are clearly one thing.

[...]

PR #240, opened 2016-02-24
Implements Help command and adds GmCommand Descriptions

Seeker1437:
* Ensure scripts aren't broken as a result of the change as well!


src/ChannelServer/Util/GmCommands.cs
```diff
@@ -35,71 +36,72 @@ public class GmCommandManager : CommandManager<GmCommand, GmCommandFunc>
...
-			Add(01, 50, "haircolor", "<hex color>", HandleHairColor);
-			Add(01, 50, "die", "", HandleDie);
-			Add(01, 50, "who", "", HandleWho);
-			Add(01, 50, "motion", "<category> <motion>", HandleMotion);
-			Add(01, 50, "gesture", "<gesture>", HandleGesture);
-			Add(01, 50, "lasttown", "", HandleLastTown);
-			Add(01, 50, "cutscene", "<name>", HandleCutscene);
-			Add(01, 50, "openshop", "<name>", HandleOpenShop);
-			Add(01, 50, "nccolor", "<id (0-32)>", HandleNameChatColor);
+			Add(01, 50, "go", "Warps to pre-defined locations", "<location>", HandleGo);
```
exectails: No localization.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -130,11 +132,12 @@ public void Add(int auth, string name, string usage, GmCommandFunc func)
 		/// <param name="auth"></param>
 		/// <param name="charAuth"></param>
 		/// <param name="name"></param>
+		/// <param name="description"></param>
 		/// <param name="usage"></param>
 		/// <param name="func"></param>
-		public void Add(int auth, int charAuth, string name, string usage, GmCommandFunc func)
+		public void Add(int auth, int charAuth, string name, string description, string usage, GmCommandFunc func)
```
exectails: Usage should go before the description, that's way more important for someone reading the code.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,114 @@ public bool Process(ChannelClient client, Creature creature, string message)
 
 		// ------------------------------------------------------------------
 
+		private CommandResult HandleHelp(ChannelClient client, Creature sender, Creature target, string message, IList<string> args)
+		{
+			var user = sender == target ? target : sender;
```
exectails: The target is always the target creature, which is the sender if no target was supplied. This line is pointless.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,114 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+			var availableCommands = _commands.Where(c => c.Value.Auth <= user.Client.Account.Authority)
+				.OrderBy(x => x.Value.Auth).ThenBy(x => x.Key).ToList();
+
+			if (args.Count < 2)
+			{
+				var commandGroups = new Dictionary<string, List<KeyValuePair<string, GmCommand>>>()
```
exectails: What if I have a server with different conventions? What if I don't have GMs that start at level 50, or a user isn't limited to < 50? And if you do want to stick to Aura conventions, where's the VIP group? This grouping system doesn't work.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,114 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+					if (authValue < 50)
+						commandGroups["User"].Add(command);
+					if (authValue >= 50 && authValue < 99)
+						commandGroups["GM"].Add(command);
+					if (authValue >= 99)
+						commandGroups["Admin"].Add(command);
+				}
+
+				Send.WhisperChat(user, "help", string.Format("There are {0} commands available to you.", availableCommands.Count));
```
exectails: Why are you using WhisperChat?
Missing localization.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,114 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+				foreach (var group in commandGroups)
+				{
+					var value = string.Join(", ", group.Value.Select(x => x.Key));
+
+					if (string.IsNullOrWhiteSpace(value)) continue;
+
+					value = value.Replace("  ", " ");
```
exectails: You have to document things that aren't obvious, what does this do? It looks way too complicated for something as simple as listing a few commands.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+				foreach (var command in availableCommands)
+				{
+					if (!commandGroups.ContainsKey(command.Value.Auth))
+						commandGroups.Add(command.Value.Auth, new List<KeyValuePair<string, GmCommand>>());
+
+					commandGroups[command.Value.Auth].Add(command);
+				}
+
+				Send.SystemMessage(target, Localization.Get("There are {0} commands available to you."), availableCommands.Count);
```
exectails: GetPlural.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+				foreach (var group in commandGroups)
+				{
+					// Joins any availabe commands into a comma separated string
+					var value = string.Join(", ", group.Value.Select(x => x.Key));
+
+					if (string.IsNullOrWhiteSpace(value)) continue;
+
+					value = value.Replace("  ", " ");
```
exectails: Where do these extra spaces come from?

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+					var words = value.Split(' ');
+					var sb = new StringBuilder();
+					var currString = new StringBuilder();
```
exectails: Why not a counter instead of an extra StringBuilder?

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+						{
+							currString.Clear();
+							sb.AppendFormat("{0}{1}", Environment.NewLine, word);
+							currString.AppendFormat(" {0}", word);
+						}
+					}
+
+					// We then break this string into a list of lines so that we can send each 
+					//     line separately
+					var blockList = sb.ToString().Split(new[] { Environment.NewLine }, StringSplitOptions.None).ToList();
```
exectails: Why not a list of strings instead of splitting a string?

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+					Send.SystemMessage(target, Localization.Get("Auth Level {0}"), group.Key);
+					Send.SystemMessage(target, "===========");
+					foreach (var line in blockList)
+					{
+						Send.SystemMessage(target, line);
+					}
+
+					Send.SystemMessage(target, "-");
```
exectails: What are these for?

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+				var commandKeyValuePair = availableCommands.FirstOrDefault(x => x.Key == targetCommand);
+				if (commandKeyValuePair.Value == null)
+				{
+					Send.SystemMessage(target, Localization.Get("Command unknown or not available to you. Try 'help' for a list of available commands."));
+					return CommandResult.Okay;
+				}
+
+				var command = commandKeyValuePair.Value;
+
+				string usableOnOthers = Localization.Get(target.Client.Account.Authority >= command.CharAuth && command.CharAuth != -1 ? "Yes" : "No");
```
exectails: This doesn't work.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+				string aliases = "";
+
+				var aliasList = availableCommands.Where(x => x.Value == command).ToList();
```
exectails: You're basically doing the same thing twice.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+				if (aliasList.Count > 1)
+				{
+					aliases = aliasList.Where(keyValuePair => keyValuePair.Key != commandKeyValuePair.Key).Aggregate(aliases, (current, keyValuePair) => current + (keyValuePair.Key + " "));
+				}
+
+				Send.SystemMessage(target, Localization.Get("Name: {0}"), commandKeyValuePair.Key);
+				if (!string.IsNullOrWhiteSpace(aliases))
+					Send.SystemMessage(target, Localization.Get("Alias(es): {0}"), aliases);
```
exectails: GetPlural.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -237,6 +240,107 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+		{
+			var availableCommands = _commands.Where(c => c.Value.Auth <= target.Client.Account.Authority)
+				.OrderBy(x => x.Value.Auth).ThenBy(x => x.Key).ToList();
+
+			if (args.Count < 2)
+			{
+				// Each group of commands is based off the auth
+				var commandGroups = new Dictionary<int, List<KeyValuePair<string, GmCommand>>>();
+
+				foreach (var command in availableCommands)
```
exectails: Is there really a use for categorizing them by auth level? That entirely depends on what the admin set, commands that are usually 0 could be 99, commands that are normally 50 could be 0, and players would look for them in the wrong categories to see if they're available.

Seeker1437:
Sorry about that... dropped my phone and accidentally hit close... didn't eve ask if I was sure... using Octodroid. xD

src/ChannelServer/Util/GmCommands.cs
Seeker1437: The only thing I could think of that would be better is possibly grouping alphanumerically.

exectails: I don't see a reason why it should be grouped.

Seeker1437:
Awesome, I have something in mind. I'll have it for you later tonight.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -238,6 +241,54 @@ public bool Process(ChannelClient client, Creature creature, string message)
...
+
+			var helpFormat = Localization.Get("{0} - {1}");
+
+			if (args.Count < 2)
+			{ 
+				Send.SystemMessage(target, Localization.GetPlural("There is {0} command available to you.", "There are {0} commands available to you.", availableCommands.Count), availableCommands.Count);
+
+				foreach (var commandKeyValuePair in availableCommands)
+				{
+					Send.SystemMessage(target, helpFormat, commandKeyValuePair.Key, commandKeyValuePair.Value.Description);
```
exectails: Looks like it could be confusing, a text wall. We already have like 60 commands and it's just gonna get more. You also might have to worry about when the chat might not display the first messages anymore. I don't know what the limit is on that, but I know there is one.

src/ChannelServer/Util/GmCommands.cs
```diff
@@ -238,6 +241,54 @@ public bool Process(ChannelClient client, Creature creature, string message)
 
 		// ------------------------------------------------------------------
 
+		private CommandResult HandleHelp(ChannelClient client, Creature sender, Creature target, string message, IList<string> args)
+		{
+			var availableCommands = _commands.Where(c => c.Value.Auth <= target.Client.Account.Authority)
+				.OrderBy(x => x.Key).ToList();
+
+			var helpFormat = Localization.Get("{0} - {1}");
```
exectails: Why localize this?

src/ChannelServer/Util/GmCommands.cs
Seeker1437: Sorry about that, when I was testing out changes before commiting, I did have a localizable string l, forgot to remove that.

I also realized I can move this inside the if block too. 

Both changes coming shortly.

exectails:
Pages... so now I have to skim through potentially dozens of pages to see what commands there are? I'm not sure about this. And what if someone adds a command named "p:1"? Maybe we should discuss how to best do this in #109 first, I'm going to close this for now.

Issue #241, opened 2016-02-24
One-string intros


exectails:
All intros of all NPCs should be turned into only one string, for easier localization. Not only requires it less `L` calls, I imagine it would be kind of irritating, encountering half sentences in the generated localization files.

Right now `Intro` takes a variable amount of string parameters, that make up the lines, that are concatenated behind the scenes. Instead, we want one string, with the lines being separated with `<br/>`s.

**Example**
Dilys' intro would be changed from this

```
await Intro(
    "A tall, slim lady tinkers with various ointments, herbs, and bandages.",
    "She looks wise beyond her years, but it might just be the healer's dress",
    "and neatly combed hair."
);
```

to this

```
await Intro("A tall, slim lady tinkers with various ointments, herbs, and bandages.<br/>She looks wise beyond her years, but it might just be the healer's dress<br/>and neatly combed hair.");
```

xCryptic:
Is this something that you want done now? I've got one more Tir NPC to do and then I can go back and change all the rest of the NPCs if you want.

I was already planning to go through all the NPCs and do the following
• check for base in class name
• add gift weights
• other savior title response (there's another one, forgot which)
• change npc name to <npcname/>

I can just add that to this list.

exectails:
@xCryptic Sounds good :+1:

exectails:
Just two more!

exectails:
:+1:

PR #246, opened 2016-02-27
Localization+XML support for NPCs

Arufonsu:
- [x] Nao, Tin and Alby's "Rescue Resident"
- [x] Tir
- [x] Sidhe
- [x] Dugald
- [x] Dunbarton
- [x] Gairech
- [x] Bangor
- [x] Tir Na Nog
- [x] Egos 

(Once the NPCs in the following list are adapted to localization and xml changes, we are supposed to create further NPCs by following these changes.)

- One-string intros ->  #241

- Localize a string
`Msg("... string ...");` -> `Msg(L("... Localized string ..."));`

- Images
`<image name='{0}' local='true' width='{0}' height='{0}' />`

- Buttons
`<button title='{0}' keyword='{1}' />`

- List
`<listbox page_size='{0}' title='{1}' cancel='{2}'> </listbox>`

- List with buttons
`<listbox page_size='2' title='list with 2 buttons' cancel='@end'><button title='button1' keyword='@1' /><button title='button2' keyword='@2' /></listbox>`

- Inputbox
`<inputbox title='{0}' caption='{1}' max_len='{2}' allow_cancel='{3}' />`


system/scripts/npcs/dungeonlostresident.cs
```diff
@@ -16,27 +16,27 @@ public override void Load()
...
-				Msg("Please take this, may it bring you luck.");
+				Msg(L("Please take this, may it bring you luck."));
 				GiveItem(16009);
 				GiveKeyword("TirChonaill_Tutorial_Perceiving");
 			}
 		}
 
 		GiveKeyword("Clear_Tutorial_Alby_Dungeon");
-		Msg("Thank you so much, now let's leave this horrible place...", Button("End Conversation"));
+		Msg(L("Thank you so much, now let's leave this horrible place...<button title='End Conversation.'/>"));
```
exectails: I *think* you always need a keyword, iirc a button doesn't work properly without. You might want to test this. In the case of End Conversation, you could use `@end`.

system/scripts/npcs/nao.cs
```diff
@@ -42,45 +38,45 @@ public override void Load()
 
 	private async Task Introduction()
 	{
-		Msg("Hello, there... You are <username/>, right?<br/>I have been waiting for you.<br/>It's good to see a " + (Player.IsMale ? "gentleman" : "lady") + " like you here.");
-		Msg("My name is Nao.<br/>It is my duty to lead pure souls like yours to Erinn.");
+		Msg(string.Format(L("Hello, there... You are <username/>, right?<br/>I have been waiting for you.<br/>It's good to see a {0} like you here."), (Player.IsMale ? L("gentleman") : L("lady"))));
```
exectails: It would probably be better to make this an if else, with two messages. Just in case this gentleman/lady thing works differently in other languages.

```csharp
if (Player.IsMale)
    Msg(L("... gentleman ..."));
else
    Msg(L("... lady ..."));
```

system/scripts/npcs/nao.cs
```diff
@@ -42,45 +38,45 @@ public override void Load()
...
 			Msg(RandomPhrase(),
-				Button("End Conversation", "@endconv"),
-				List("Talk to Nao", 4, "@endconv",
-					Button("About Mabinogi", "@mabinogi"),
-					Button("About Erinn", "@erinn"),
-					Button("What to do?", "@what"),
-					Button("About Adventures", "@adventures")
+				Button(L("End Conversation"), "@endconv"),
+				List(L("Talk to Nao"), 4, "@endconv",
+					Button(L("About Mabinogi"), "@mabinogi"),
```
exectails: The purpose of the XML issue is to get rid of all dialog elements, this includes `List`. Though I just saw there are more buttons and images later, you didn't convert everything yet, did you?
https://github.com/aura-project/aura/blob/master/src/ChannelServer/Scripting/Scripts/DialogElements.cs#L252

system/scripts/npcs/nao.cs
```diff
@@ -155,13 +150,7 @@ private string RandomPhrase()
 							for (int i = 1; i < 10; ++i)
 								RemoveKeyword("Tin_ColorAmpul_" + i);
 							RemoveKeyword("tutorial_present");
-
-							// Old:
```
exectails: You shouldn't remove comments, these could be important.

system/scripts/npcs/tir/alissa.cs
```diff
@@ -283,8 +278,8 @@ protected void BuyWindmill(int gold, int minutes)
 		ActivateWindmill(minutes);
 
 		RndMsg(
-			"Okay!<br/>Anyone can use the Mill now for the next " + minutes + " minutes.<br/>I'm counting, haha.",
-			"Yay! I got some pocket money!"
+			L("Okay!<br/>Anyone can use the Mill now for the next ") + minutes + L(" minutes.<br/>I'm counting, haha."),
```
exectails: This would be a job for `string.Format`. And I guess we need plural version for `L`...

system/scripts/npcs/tir/caitin.cs
```diff
@@ -267,153 +263,153 @@ public class CaitinShop : NpcShopScript
...
+		Add(L("Grocery"), 50430, 1);  // Grapes x1
+		Add(L("Grocery"), 50430, 5);  // Grapes x5
+		Add(L("Grocery"), 50430, 10); // Grapes x10
+		Add(L("Grocery"), 50430, 20); // Grapes x20
+		Add(L("Grocery"), 50431, 1);  // Ripe Pumpkin x1
+		Add(L("Grocery"), 50431, 5);  // Ripe Pumpkin x5
+		Add(L("Grocery"), 50431, 10); // Ripe Pumpkin x10
+		Add(L("Grocery"), 50431, 20); // Ripe Pumpkin x20
+
+		Add(L("Gift"), 52010); // Ramen
```
exectails: I wonder if we should use contexts for such short strings, the meaning could be different, depending on the circumstances.

system/scripts/npcs/nao.cs
Arufonsu: Oh, that is very helpful now! please tell me if i've mistaken but this should go like this right?...

```
			Msg(RandomPhrase(),
					L("<button title='End Conversation' keyword='@endconv' />"),
					L("<listbox page_size='4' title='Talk to Nao' cancel='@endconv'><button title='About Mabinogi' keyword='@mabinogi' /><button title='About Erinn' keyword='@erinn' /><button title='What to do?' keyword='@what' /><button title='About Adventures' keyword='@adventures' /></listbox>")
			);
```

This worked fine so far

system/scripts/npcs/nao.cs
Arufonsu: My mistake, already fixed.

system/scripts/npcs/dungeonlostresident.cs
Arufonsu: hmm you are right, so just in case of being necessary, already added `@end`

exectails:
I'm thinking, maybe we don't want to get rid of the dialog system altogether, or rather, the methods.

For example, the list code is rather long and error prone. One wrong character in the code or the localization and the dialog breaks. And if you aren't writing these codes for a living, you might never remember all the required and optional parameters, default values, etc. Hell, I have trouble remember the *names* of the tags... What does `face` do? What's the code for hiding the NPC face?

This
```csharp
List(L("list with 2 buttons"), Button(L("button1"), "@1"), Button(L("button2"), "@2"))
```
is much easier to read, write, and remember than
```csharp
L("<listbox page_size='2' title='list with 2 buttons' cancel='@end'><button title='button1' keyword='@1' /><button title='button2' keyword='@2' /></listbox>")
```

Same goes for other codes, even if they're simpler.

Arufonsu:
Alright.. Lets see if we are on the same page: 

1) I would revert back (from the changes i have done so far) the following methods:

-Buttons
Why? An example should do it.. Most NPCs have the "Start a conversation" or "Start Conversation" button right? Well, Poedit will grab all of those buttons and make it so you have to translate that button's string only ONCE, same situation applies to shop's tabs.

-Inputs (same reason i have for buttons, most inputs have the same string)

-Images (don't really needed for localization?)

-Face Expression (same reason i have for images, also is rarely used, have seen it on egos so far, maybe few quests.)

-Lists


2) I would keep:

-Localization, obviously.

-One line intro (better for localization porpouses, doesnt really affects the coding task..)

-String method / Singular and Plurals for repairs/mill/etc

exectails:
Yea, pretty much. But I'm not sure about that yet, just wanted to put it up for discussion.

exectails:
>you have to translate that button's string only ONCE

This is an especially nice side-effect.

PR #249, opened 2016-03-04
support MySQL 5.7

phstudy:


sql/main.sql
```diff
@@ -1,4 +1,4 @@
-CREATE DATABASE `aura` DEFAULT CHARACTER SET utf8 COLLATE utf8_general_ci;
+CREATE DATABASE IF NOT EXISTS `aura` DEFAULT CHARACTER SET utf8 COLLATE utf8_general_ci;
```
QewQew: This can cause potential issues.
What if I already had a database called aura (not for aura) prior to setting up aura?
I don't want new tables added to that database, then.

sql/main.sql
```diff
@@ -107,7 +107,7 @@ CREATE TABLE IF NOT EXISTS `creatures` (
 ) ENGINE=InnoDB  DEFAULT CHARSET=utf8 AUTO_INCREMENT=2 ;
 
 INSERT INTO `creatures` (`creatureId`, `server`, `name`, `race`, `skinColor`, `eyeType`, `eyeColor`, `mouthType`, `height`, `weight`, `upper`, `lower`, `color1`, `color2`, `color3`, `region`, `x`, `y`, `direction`, `level`, `levelTotal`, `exp`, `ap`, `age`, `lifeMax`, `lifeDelta`, `injuries`, `manaMax`, `manaDelta`, `staminaMax`, `staminaDelta`, `hunger`, `str`, `int`, `dex`, `will`, `luck`, `lifeFood`, `manaFood`, `staminaFood`, `strFood`, `intFood`, `dexFood`, `willFood`, `luckFood`, `defense`, `protection`, `deletionTime`, `weaponSet`, `title`, `optionTitle`, `state`, `creationTime`, `lastAging`, `lastRebirth`, `lastLogin`) VALUES
-(1, 'AuraSystem', '_Dummy', 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 1, 12800, 38100, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, NULL, 0, 0, 0, 0, '0000-00-00 00:00:00', '0000-00-00 00:00:00', NULL, NULL);
+(1, 'AuraSystem', '_Dummy', 0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 1, 12800, 38100, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, NULL, 0, 0, 0, 0, '1970-01-01 00:00:01', '1970-01-01 00:00:01', NULL, NULL);
```
QewQew: Why does this need to be changed in order to support MySQL 5.7?

phstudy: MySQL 5.7 disallows date '0000-00-00 by default. 

http://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sql-mode-changes

sql/main.sql
phstudy: This makes sense.

sql/main.sql
exectails: That page also mentions that the change was reverted for now, but I see the point^^

PR #250, opened 2016-03-05
Updated Piaras

xCryptic:
• Added greetings
• Added missing keywords
• Other small fixes

The skills/campfire keywords seem to match. I tested them and got the results you see in the code. I guess that means it was changed not too long ago.

### DIFF
--- a/system/scripts/npcs/tir/piaras.cs
+++ b/system/scripts/npcs/tir/piaras.cs
@@ -83,20 +83,16 @@ public class PiarasScript : NpcScript
 				break;
 
 			case "about_skill":
-				if (Player.IsHuman)
+				// When the beginner quests changed, he no longer removes the keyword. Instead he says "I'm sorry.<br/>I don't have the time to talk about that right now."
+				// He also checks for race, but until we get the respective quests in for other races, we'll leave it out
+				if (Player.Skills.Has(SkillId.Campfire))
 				{
-					if (Player.Skills.Has(SkillId.Campfire))
-					{
-						RemoveKeyword("skill_campfire");
-						Msg("Ha ha. Now you know how to use the Campfire skill.<br/>It's something I didn't want to teach you, to be honest,<br/>but I am impressed that you have mastered it so well.<br/>With this, another young adventurer is born today, ha ha.");
-					}
-					else
-					{
-						Msg("I'm sorry.<br/>I don't have the time to talk about that right now.");
-					}
+					RemoveKeyword("skill_campfire");
+					Msg("Ha ha. Now you know how to use the Campfire skill.<br/>It's something I didn't want to teach you, to be honest,<br/>but I am impressed that you have mastered it so well.<br/>With this, another young adventurer is born today, ha ha.");
 				}
 				else
 				{
+					GiveKeyword("skill_campfire");
 					Msg("Do you by chance know about the Campfire Skill?");
 					Msg("If you start a fire using the Campfire Skill,<br/>people would come by one at a time after seeing the bright fire from afar...");
 					Msg("People share what they have in their inventory<br/>and spend long summer nights sharing stories about their adventures.");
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tir/piaras.cs
```diff
@@ -60,24 +72,31 @@ public override void Load()
...
 
 			case "rumor":
+				GiveKeyword("square");
 				Msg("Why don't you talk to others in town? There's a good spot to meet people. The Town Square is right up this way. I suggest you try there first.");
 				ModifyRelation(Random(2), 0, Random(2));
 				break;
 
 			case "about_skill":
-				if (Player.Skills.Has(SkillId.Campfire))
+				if (Player.IsHuman)
```
exectails: Adding such a check is a bad idea. It will take some time until we get to G4+ and people *will* create elves and giants before that. Such checks restrict what they can do.

system/scripts/npcs/tir/piaras.cs
```diff
@@ -60,24 +72,31 @@ public override void Load()
...
+						Msg("Ha ha. Now you know how to use the Campfire skill.<br/>It's something I didn't want to teach you, to be honest,<br/>but I am impressed that you have mastered it so well.<br/>With this, another young adventurer is born today, ha ha.");
+					}
+					else
+					{
+						Msg("I'm sorry.<br/>I don't have the time to talk about that right now.");
+					}
 				}
 				else
 				{
-					GiveKeyword("skill_campfire");
```
exectails: Why did you remove this? I (probably) added it for a reason.

xCryptic: Because I didn't get it on official. I went on my elf without the keyword and didn't obtain it.

system/scripts/npcs/tir/piaras.cs
xCryptic: Well yeah, but how is it restricting? It's just different dialogue, not anything game-breaking. I think it should stay to keep it as official-like as possible.

system/scripts/npcs/tir/piaras.cs
Arufonsu: "Piaras used to give it in response to [Skills], but now the keyword is not given."

http://wiki.mabinogiworld.com/view/Campfire_Skill_Keyword

system/scripts/npcs/tir/piaras.cs
exectails: The campfire quest *depends* on this dialog giving and removing the proper keywords.

system/scripts/npcs/tir/piaras.cs
exectails: See what I said above, and check the beginner quest "202026_campfire".

xCryptic: Wait... but then- what? Then it works different on official? Or are you using the old beginner quests? (which would mean this should be a feature check)

exectails: Of course I'm using the old quests o,o This quest doesn't even exist on NA anymore I think. It probably should have a feature check, but should we add feature checks left and right for every little thing already? Or should we wait until these features actually do something? In this case that would also require some research, as to when this quest and the keyword vanished.

xCryptic: Ohh I see. I guess for now I'll comment out the race check and leave a comment about it being removed when the beginner quests changed (assuming that's when it was).

system/scripts/npcs/tir/piaras.cs
```diff
@@ -159,12 +229,45 @@ public override void Load()
...
+				// When the beginner quests changed, he no longer removes the keyword. Instead he says "I'm sorry.<br/>I don't have the time to talk about that right now."
+				// He also checks for race, but until we get the respective quests in for other races, we'll leave it out
+				if (Player.Skills.Has(SkillId.Campfire))
+				{
+					RemoveKeyword("skill_campfire");
+					Msg("Ha ha. Now you know how to use the Campfire skill.<br/>It's something I didn't want to teach you, to be honest,<br/>but I am impressed that you have mastered it so well.<br/>With this, another young adventurer is born today, ha ha.");
+				}
+				else
+				{
+					GiveKeyword("skill_campfire");
```
exectails: Why would he give you the keyword when you talk to him *with said keyword*? C&P error? Was the dialog you had before official, or did you just copy it as well?

xCryptic: Oh oops. That was just C&P. He still says the same thing as the skills keyword though.

PR #252, opened 2016-03-14
(Scripts) Localization update

Arufonsu:
- Hunting quests
- Collection quests
- Party quests
- Scroll quests


exectails:
ss 2016-03-15 at 12 03 47
We have 165 quests? Crazy @_@"

Arufonsu:
> We have 165 quests? Crazy @_@" Gj^^

Oh.. thanks ^^, and no, there's 37 more quests we need to localize, but they are part of the methods we have been discussing these days.... so for now aura has 202 working quests and 165 of them with localization support :dart:

PR #255, opened 2016-03-19
Fixes #251

### LINKED ISSUE #251: No stamina consumption when attacking props
Attacking props (trees, bushes, poles, etc) doesn't reduces stamina. In original, they consume stamina in the same way a combat attack does.

Tested both with weapons and bare-hands.

Searched for a similar issue but couldn't find any. #125 Is the closer I've got, seems that applying the same stamina calculation for props would do. I'll send a patch If I manage to fix it.

gcscaglia:
* Refactor dual wielding flags from ``CombatMastery.cs`` to ``Creature.cs``
* Fixes wrong stun for prop hits while dual wielding.
* Include stamina consumption into ``Prop.cs``

### DIFF
--- a/src/ChannelServer/Network/Handlers/Props.cs
+++ b/src/ChannelServer/Network/Handlers/Props.cs
@@ -48,24 +48,36 @@ namespace Aura.Channel.Network.Handlers
 			}
 
 			// Check range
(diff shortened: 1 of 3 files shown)


exectails:
Nice job, thanks.

PR #256, opened 2016-03-26
Added GM command description support to GeneralScript

xeroplz:
Adding commands with descriptions was left out of GeneralScript, so here you go.


exectails:
Oh, totally forgot about that, thanks.

PR #272, opened 2016-05-14
Updated Trefor

xCryptic:
• Added missing keywords
• Added greetings
• Added upgrading
• Added missing BGM
• Other small fixes

### DIFF
--- a/system/scripts/npcs/tir/trefor.cs
+++ b/system/scripts/npcs/tir/trefor.cs
@@ -224,8 +224,7 @@ public class TreforScript : NpcScript
 				Msg("This skill, however, requires a high level of concentration<br/>in that you'll have to accurately anticipate your enemy's next move.<br/>The main drawbacks of using this skill is that your Stamina will be continually spent,<br/>and you won't be able to move during the skill.");
 				Msg("Now let's see your stance.<br/>Oh no... What kind of a stance is that? You won't be able to react quick enough with that.<br/>Actually, you won't be able to fight a raccoon with that.");
 				Msg("The basics of the Counterattack skill involves utilizing your enemy's power.<br/>This skill does not involve utilizing YOUR strength.");
-				Msg("Your legs are all tensed up. You can't react in time like that.<br/>Loosen your left leg to make sure you can absorb your enemy's strength...<br/>Yes. That's it. Now you look like you're ready.", Button("Continue", "@yes"));
-				await Select();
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tir/trefor.cs
```diff
@@ -156,24 +198,52 @@ public override void Load()
...
 				Msg("This skill, however, requires a high level of concentration<br/>in that you'll have to accurately anticipate your enemy's next move.<br/>The main drawbacks of using this skill is that your Stamina will be continually spent,<br/>and you won't be able to move during the skill.");
 				Msg("Now let's see your stance.<br/>Oh no... What kind of a stance is that? You won't be able to react quick enough with that.<br/>Actually, you won't be able to fight a raccoon with that.");
 				Msg("The basics of the Counterattack skill involves utilizing your enemy's power.<br/>This skill does not involve utilizing YOUR strength.");
-				Msg("Your legs are all tensed up. You can't react in time like that.<br/>Loosen your left leg to make sure you can absorb your enemy's strength...<br/>Yes. That's it. Now you look like you're ready.");
+				Msg("Your legs are all tensed up. You can't react in time like that.<br/>Loosen your left leg to make sure you can absorb your enemy's strength...<br/>Yes. That's it. Now you look like you're ready.", Button("Continue", "@yes"));
```
exectails: Why did you add an explicit Continue button? o.o

xCryptic: That's how it is. I match everything exactly how it's logged. (unless there's grammar errors)

exectails: Feel free to change things if they don't many *any* sense o,o We don't want to copy Mabi 1:1, we only want the correct *behavior*^^

system/scripts/npcs/tir/trefor.cs
```diff
@@ -39,6 +42,10 @@ public override void Load()
 
 	protected override async Task Talk()
 	{
+		SetBgm("NPC_Guard.mp3");
+
+		// I noticed the intro message is different as of r226 on 5/11/16
+		// "A specimen of physical fitness stands at attention in a suit of<br/>immaculate armor. Through his lowered visor, you catch the<br/>slightest flash of his determined eyes."
```
exectails: I wonder if we should feature check those... not like it's an earth-shattering difference, but a difference nonetheless.

PR #306, opened 2016-06-14
Added "a Friend of X" Title Gifts

xCryptic:


system/scripts/npcs/tir/duncan.cs
```diff
@@ -55,33 +55,49 @@ public override void Load()
...
+							break;
+
+						case 10062: // is a friend of Nora
+							npc = L("Nora");
+							break;
+					}
+
+					if (npc != null)
+					{
+						Msg(L("That's great, <username/>.<br/>Seeing you and " + npc + " are such good friends<br/>makes me feel great as the chief of this town."));
```
exectails: This doesn't work with localization, you have to use string.Format.

system/scripts/npcs/tir/caitin.cs
```diff
@@ -49,8 +49,26 @@ public override void Load()
...
+					{
+						Msg(L("Nice to meet you, <username/>. I heard many things about you from Malcolm."));
+						Player.Vars.Perm["caitin_title_gift"] = today;
+						GiveItem(50134); // Sliced Bread
+						Notice(L("Received Sliced Bread from Caitin."));
+						SystemMsg(L("Received Sliced Bread from Caitin."));
+						Msg(L("Would you like to have a taste of this?<br/>It's nothing special...but I made it myself."));
+					}
+				}
+				else if (Player.Titles.SelectedTitle == 11002)
```
exectails: On the others you didn't do else if for this title, is it actually different for Caitin?

system/scripts/npcs/tir/trefor.cs
```diff
@@ -60,13 +60,43 @@ public override void Load()
 			case "@talk":
 				Greet();
 				Msg(Hide.Name, GetMoodString(), FavorExpression());
-				if (Player.Titles.SelectedTitle == 11002)
+
+				var playerVar = Player.Vars.Perm["trefor_title_gift"];
+				var today = ErinnTime.Now.ToString("yyyyMMdd");
+				string message = null;
+
+				if (playerVar != today || playerVar == null)
```
exectails: This should actually work fine with just the `!=`. Since the return values of Vars are [dynamic](https://msdn.microsoft.com/en-us/library/dd264741.aspx), the framework will figure out the check at run-time.

exectails:
It would've been better if you had PRed for the title branch.

system/scripts/npcs/tir/caitin.cs
xCryptic: Ugh. I went through my changes 3 times, and somehow I missed this (forgot to change selectedtitle to just Title). There's no difference between the NPCs, but does it matter whether it's an else if or just another if?

system/scripts/npcs/tir/trefor.cs
xCryptic: Really? When I was working on my halloween event, I wasn't checking the playerVar for null, and it was throwing errors back then. I didn't try it recently, but I can go back and test it now.

xCryptic:
Oh, sorry. I totally forgot you were working on a different branch for this. Why did you merge it if there's still issues?

system/scripts/npcs/tir/trefor.cs
exectails: In your event you tried to access a property on the dynamic object, which would throw an exception if the object was null.

system/scripts/npcs/tir/caitin.cs
exectails: Now that I think about it, it obviously doesn't matter^^" Now I see why you did it like this in a few NPCs, because those were the ones were you only checked for the title in this scope. Nvm.

exectails:
So I could fix the issues and close the issue^^

PR #333, opened 2016-06-29
Dual Gun Skills added

xeroplz:
- Adds all Dual Gun skills
- Adds bulletsMax entries in the item database for guns
- Adds partial support for MultiHit TargetOptions


exectails:
So many skills in one PR q________q

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+			if (!creature.RightHand.MetaData1.Has(BulletCountTag))
+			{
+				creature.RightHand.MetaData1.SetShort(BulletCountTag, 0);
+				Send.ItemUpdate(creature, creature.RightHand);
+			}
+
+			// Check Bullet Count
+			var bulletCount = creature.RightHand.MetaData1.GetShort(BulletCountTag);
+			if (bulletCount < skill.RankData.Var1 && !creature.Conditions.Has(ConditionsD.WayOfTheGun))
+				Send.SkillPrepareSilentCancel(creature, skill.Info.Id);
```
exectails: First you send silent cancel, then you send prepare? Is this intended?

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+			target.StopMove();
+
+			// Slide
+			var targetPos = target.GetPosition();
+			attacker.TurnTo(targetPos);
+			var attackerPos = attacker.GetPosition();
+			var newAttackerPos = targetPos.GetRelative(attackerPos, SlideDistance);
+
+			// Set new attacker position to nearest spot if there is a collision
+			if (attacker.Region.Collisions.Any(attackerPos, newAttackerPos))
```
exectails: You're basically checking for collisions twice by first doing Any and then Find, this is unnecessary work. You should use only Find. From AiScript:
```csharp
// Check for collision
Position intersection;
if (this.Creature.Region.Collisions.Find(pos, destination, out intersection))
	destination = pos.GetRelative(intersection, -100);
```
Any is a convenience method for when you don't care where exactly the collision is.

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+			{
+				Position intersection;
+				attacker.Region.Collisions.Find(attackerPos, newAttackerPos, out intersection);
+				newAttackerPos = attackerPos.GetRelative(intersection, -50);
+			}
+
+			Send.ForceRunTo(attacker, newAttackerPos);
```
exectails: Are you setting the creature's position anywhere? If you only send a force run, but don't actually move the creature on the server, you're "unsyncing" them.

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+
+				// Always Knock Back
+				if (target.Is(RaceStands.KnockBackable))
+				{
+					tAction.Set(TargetOptions.KnockBack);
+					attacker.Shove(target, KnockbackDistance);
+				}
+				tAction.Creature.Stun = tAction.Stun;
+			}
+			aAction.Creature.Stun = aAction.Stun;
```
exectails: CaombatActionPack handles this automatically, you only have to set the stun on the action.

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+			cap.Handle();
+
+			// Item Update excluding Way Of The Gun
+			if (!attacker.Conditions.Has(ConditionsD.WayOfTheGun))
+			{
+				var bulletCount = attacker.RightHand.MetaData1.GetShort(BulletCountTag);
```
exectails: Under the assumption that I'll find something like this in every one of those skills, you should really make this a helpe method. Maybe in SkillHelper.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+		/// Prepares the skill
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		/// <returns></returns>
+		public bool Prepare(Creature creature, Skill skill, Packet packet)
+		{
+			if (creature.RightHand == null)
+				Send.SkillPrepareSilentCancel(creature, skill.Info.Id);
```
exectails: Again you're silently canceling, but then you continue like nothing happened, I can't imagine this being intended =| Silent cancel is sent automatically if you return false from Prepare.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+
+			// Set Bullet Count Tag if it doesn't exist
+			if (!creature.RightHand.MetaData1.Has(BulletCountTag))
```
exectails: Probably another thing that could use a helper method.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+			{
+				creature.RightHand.MetaData1.SetShort(BulletCountTag, 0);
+				Send.ItemUpdate(creature, creature.RightHand);
+			}
+
+			// Check Bullet Count
+			var bulletCount = creature.RightHand.MetaData1.GetShort(BulletCountTag);
+			if (bulletCount < skill.RankData.Var1 && !creature.Conditions.Has(ConditionsD.WayOfTheGun))
+				Send.SkillPrepareSilentCancel(creature, skill.Info.Id);
```
exectails: Last shout out for canceling without returning.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Use(Creature attacker, Skill skill, Packet packet)
+		{
+			var targetAreaId = packet.GetLong();
+			var targetAreaLoc = new Location(targetAreaId);
+			var targetAreaPos = new Position(targetAreaLoc.X, targetAreaLoc.Y);
+
+			attacker.StopMove(); // Unnecessary at the moment, but who knows.
+			attacker.Lock(Locks.Walk | Locks.Run);
```
exectails: These two should get locked automatically. You don't have to add any locks/unlocks that can be found in the skill db, those are done automatically.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+			var cap = new CombatActionPack(attacker, skill.Info.Id);
+
+			var aAction = new AttackerAction(CombatActionType.SpecialHit, attacker, targetAreaId);
+			aAction.Set(AttackerOptions.UseEffect | AttackerOptions.KnockBackHit1);
+			cap.Add(aAction);
+
+			aAction.Stun = AttackerStun;
+
+			// Get targets in descending order for effect packet using bulletDist
+			var targets = attacker.Region.GetCreaturesInPolygon(p1, p2, p3, p4).Where(x => attacker.CanTarget(x) && !attacker.Region.Collisions.Any(attackerPos, x.GetPosition())).OrderByDescending(x => x.GetPosition().GetDistance(attackerPos)).Reverse().ToList();
```
exectails: First you order by DESC, then you reverse it? It would also be faster to not turn the result into a list, but only Count() the amount for later use, while iterating the IEnumerable, unless you need indexing.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+
+			// Filter by max targets [var5] and bullet count
+			var bulletCount = attacker.RightHand.MetaData1.GetShort(BulletCountTag);
```
exectails: I think counting the bullets could a method on the creature, GetBulletCount() or something. No need to have the tag constant and this line in all handlers.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+
+			if (attacker.Conditions.Has(ConditionsD.WayOfTheGun))
+				bulletCount = attacker.RightHand.MetaData1.GetShort(BulletsMaxTag);
+
+			var maxTargetsBulletCount = (int)Math.Floor((decimal)(bulletCount / 4));
+
+			if (maxTargetsBulletCount < skill.RankData.Var5)
+			{
+				targets = targets.Take(maxTargetsBulletCount).ToList();
```
exectails: If you don't use a List above, you don't need to turn these into lists either. More time saved.

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+			}
+			else if (maxTargetsBulletCount >= skill.RankData.Var5)
+			{
+				targets = targets.Take((int)skill.RankData.Var5).ToList();
+			}
+
+			// Effects & Etc
+			var shootEffect = new Packet(Op.Effect, attacker.EntityId);
```
exectails: I see why you put this Packet here, but I'd wish we'd find a way to move it to Send. After saving time removing those List creations above, maybe we can now create a list here, with target ids and distances to pass to Send.ShootEffect?^^

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,496 @@
...
+					break;
+
+				case SkillRank.R1:
+					if (targets.Length >= 6) attackerSkill.Train(4); // 6 or more enemies
+					if (targets.Length >= 10) attackerSkill.Train(5); // 10 or more enemies
+					break;
+			}
+		}
+
+		private Point RotatePoint(Point point, Point pivot, double radians)
```
exectails: We're already using this method in two skills in the core, we should probably move it somewhere for skills to use.

src/ChannelServer/Skills/Guns/DualGunMastery.cs
```diff
@@ -0,0 +1,303 @@
...
+				var cap = new CombatActionPack(attacker, skill.Info.Id);
+				cap.Type = CombatActionPackType.ChainRangeAttack;
+				cap.Hit = i;
+				cap.PrevId = prevId;
+				prevId = cap.Id;
+
+				// Prepare Combat Actions
+				var aAction = new AttackerAction(CombatActionType.RangeHit, attacker, targetEntityId);
+				aAction.Set(AttackerOptions.Result | AttackerOptions.KnockBackHit1 | AttackerOptions.KnockBackHit2);
+				aAction.SkillId = skill.Info.Id; // Dual Gun Mastery
```
exectails: Since AttackerAction already sets the SkillId to CombatMastery automatically if you don't have a skill active, it should probably set it to DualGunMastery if you have a gun equipped. That way this line isn't needed.

src/ChannelServer/Skills/Guns/BulletStorm.cs
xeroplz: So instead of sending that, I should just return false - is what you're saying?

src/ChannelServer/Skills/Guns/FlashLauncher.cs
```diff
@@ -0,0 +1,295 @@
...
+			// Get Target
+			var target = attacker.Region.GetCreature(targetEntityId);
+
+			// Check Target + Collisions
+			if (target == null || attacker.Region.Collisions.Any(attacker.GetPosition(), target.GetPosition()))
+				return CombatSkillResult.InvalidTarget;
+
+			var targetPos = target.GetPosition();
+
+			/*
```
exectails: Left over test code?

src/ChannelServer/Skills/Guns/BulletStorm.cs
xeroplz: I don't recall seeing an order by ascending, so I just had to do descending and reverse. I'll take a look at it again though.

src/ChannelServer/Skills/Guns/FlashLauncher.cs
```diff
@@ -0,0 +1,295 @@
...
+
+			// Prepare combat actions
+			var cap = new CombatActionPack(attacker, skill.Info.Id);
+
+			var aAction = new AttackerAction(CombatActionType.RangeHit, attacker, targetEntityId);
+			aAction.Set(AttackerOptions.Result | AttackerOptions.KnockBackHit1 | AttackerOptions.KnockBackHit2);
+
+			var tAction = new TargetAction(CombatActionType.TakeHit, target, attacker, SkillId.CombatMastery);
+			tAction.Set(TargetOptions.Result);
+			tAction.AttackerSkillId = skill.Info.Id; // Flash Launcher
```
exectails: Why do you pass CombatMastery above if you want the attacker skill id to be the skills's id?

src/ChannelServer/Skills/Guns/FlashLauncher.cs
```diff
@@ -0,0 +1,295 @@
...
+
+			// Damage
+			var damage = (attacker.GetRndDualGunDamage() * (skill.RankData.Var2 / 100f));
+
+			// Master Title
+			if (attacker.Titles.SelectedTitle == skill.Data.MasterTitle)
+				damage += (damage * (skill.RankData.Var3 / 100f)); // +20% damage
+
+			// Critical Hit
+			var dgm = attacker.Skills.Get(SkillId.DualGunMastery);
```
exectails: You seem to have this code in all those skills, if you need a snippet more than 1-2 times it's beneficial to make a helper for it. It makes the handlers easier to read and understand and the code reusable without having to C&P it.

src/ChannelServer/Skills/Guns/FlashLauncher.cs
```diff
@@ -0,0 +1,295 @@
...
+					target.Stability -= StabilityReduction;
+				}
+
+				// Always knock down
+				if (target.Is(RaceStands.KnockDownable))
+				{
+					tAction.Set(TargetOptions.KnockDown);
+					attacker.Shove(target, KnockbackDistance);
+				}
+				tAction.Creature.Stun = tAction.Stun;
```
exectails: Oh, the same thing I said about the attacker stun applies to the target as well of course.

src/ChannelServer/Skills/Guns/FlashLauncher.cs
```diff
@@ -0,0 +1,295 @@
...
+
+			var attackerSlidePos = attacker.GetPosition().GetRelative(targetPos, SlideDistance);
+
+			// Effects
+			Send.EffectDelayed(target, 400, Effect.FlashLauncher, (byte)0, (float)targetPos.X, (float)targetPos.Y, 400, 1000); // ?
+			Send.EffectDelayed(attacker, 1400, Effect.FlashLauncher, (byte)1, (float)attackerSlidePos.X, (float)attackerSlidePos.Y, 800); // Flash Launcher burst effect
+			Send.EffectDelayed(target, 3500, Effect.FlashLauncher, (byte)2); // ?
+
+			// Item Update excluding Way Of The Gun
+			if (!attacker.Conditions.Has(ConditionsD.WayOfTheGun))
```
exectails: .

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+				return;
+			}
+
+			// Check Range
+			if (attacker.GetPosition().InRange(target.GetPosition(), (int)skill.RankData.Var4)) // If attacker is within Max Range
+			{
+				if (!attacker.GetPosition().InRange(target.GetPosition(), (int)skill.RankData.Var5)) // If attacker isn't within Min Range
+				{
+					// Proceed
+					goto L_Proceed;
```
exectails: You don't need a goto here at all. Check range, if invalid, send notice, otherwise, continue. Check other range, if invalid, send notice, otherwise, continue. This makes it arguably easier to follow.

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+				Send.Notice(attacker, Localization.Get("You are too far away."));
+				Send.SkillUseSilentCancel(attacker);
+				attacker.Unlock(Locks.Walk | Locks.Run);
+				return;
+			}
+
+			L_Proceed:
+
+			// Position calculations
+			attacker.StopMove();
```
exectails: StopMove returns the new position, you don't need to get it again below.

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+			attacker.Lock(Locks.Walk | Locks.Run);
+			var attackerPos = attacker.GetPosition();
+			var targetPos = target.GetPosition();
```
exectails: You aren't stopping the target, are you?

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+
+				aAction.Stun = AttackerStun;
+
+				// Death or Knockback
+				if (target.IsDead)
+				{
+					tAction.Set(TargetOptions.FinishingKnockDown);
+					attacker.Shove(target, KnockbackDistance);
+
+					// cap must be handled here as targets wouldn't targetable when they're dead.
```
exectails: Doesn't this effectively mean you handle cap either way? You handle it here if they're dead, and you handle it below if they aren't dead. And if you didn't handle cap you wouldn't see any effects.

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+			this.Complete(attacker, skill, packet);
+		}
+
+		/// <summary>
+		/// Completes the skill
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Complete(Creature creature, Skill skill, Packet packet)
```
exectails: This method is misleading, since the skills is not actually completable, it's completed as the server finishes it. If you want to keep this in a separate method you should at least remove the packet parameter, so it's less ambiguous.

Also, in the logs SkillComplete is sent before the cap, does it still work as intended?

src/ChannelServer/Skills/Guns/GrappleShot.cs
```diff
@@ -0,0 +1,393 @@
...
+		{
+			Send.SkillComplete(creature, skill.Info.Id);
+			creature.Skills.ActiveSkill = null;
+
+			// Unlock Leftover Locks
+			creature.Unlock(Locks.Walk | Locks.Run | Locks.PickUpAndDrop | Locks.TalkToNpc | Locks.ChanceStance | Locks.Attack | Locks.ChangeEquipment);
```
exectails: Why not unlock everything, like the completion packet handler would?

src/ChannelServer/Skills/Guns/Reload.cs
```diff
@@ -0,0 +1,182 @@
...
+	{
+		/// <summary>
+		/// Bullet Count Tag for Gun
+		/// </summary>
+		private const string BulletCountTag = "GVBC";
+
+		/// <summary>
+		/// Bullet Min damage tag
+		/// </summary>
+		private const string GBAMINTag = "GBAMIN";
```
exectails: Maybe name them something more descriptive, otherwise you could just as well use the strings^^

src/ChannelServer/Skills/Guns/Reload.cs
```diff
@@ -0,0 +1,182 @@
...
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		/// <returns></returns>
+		public bool Ready(Creature creature, Skill skill, Packet packet)
+		{
+			creature.StopMove();
+
+			// Following Packet Data for Stacks... DevCat why?
```
exectails: The correct question is "xero why?", because you don't have to mindlessly copy them^^ Setting Stacks and then resetting it to 0 doesn't to anything, it has literally no effect on Aura, aside from those two packets that get sent. Unless the client needs this, which should be documented, you can safely remove them.

It's possible this pattern triggered something on the official server, or that two lines of code that do something else triggered those packets being sent. Doesn't mean we have to do the same =)

src/ChannelServer/Skills/Guns/Reload.cs
```diff
@@ -0,0 +1,182 @@
...
+
+				case 45038: // High-Density Mana Bullet
+					damageAdded = 20;
+					break;
+
+				case 53564: // Suntouched Mana Bullet
+					damageAdded = 30;
+					break;
+
+				default:
```
exectails: You already set it to 0, this case is redundant.

src/ChannelServer/Skills/Guns/ShootingRush.cs
```diff
@@ -0,0 +1,465 @@
...
+			var newAttackerPoint = new Point(newAttackerPos.X, newAttackerPos.Y);
+
+			var pointDist = Math.Sqrt((distance * distance) + (radius * radius)); // Pythagorean Theorem - Distance between point and opposite side's center.
+			var rotationAngle = Math.Asin(radius / pointDist);
+
+			// Note: My previous error was that I used pointDist for the GetRelative distance,
+			// which was wrong since it adds that distance to the current distance.
+			// The only extra distance I needed was that of the pointDist - distance.
+
+			// Calculate Points 1 & 2
```
exectails: This is another block you had before, right? Prime candidate for a helper.

src/ChannelServer/Skills/Guns/WayOfTheGun.cs
```diff
@@ -0,0 +1,204 @@
...
+			skill.State = SkillState.Used;
+
+			var duration = (int)skill.RankData.Var1;
+
+			// Master Title - +5 seconds
+			if (creature.Titles.SelectedTitle == skill.Data.MasterTitle)
+				duration += 5;
+
+			// Deactivate condition after skill duration
+			Task.Delay(duration).ContinueWith(_ =>
```
exectails: Delay uses milliseconds, but you're adding +5*s* above.

src/ChannelServer/World/Entities/Creature.cs
```diff
@@ -1867,6 +1867,45 @@ public float GetRndMagicDamage(Skill skill, float baseMin, float baseMax)
...
+			var totalMaxDamage = maxDamage + gunMinDamage;
+
+			var addedBulletDamage = 0;
+
+			// Max from Gun Bullet
+			if (this.RightHand != null)
+			{
+				if (this.RightHand.MetaData1.Has("GBAMAX"))
+				{
+					addedBulletDamage = this.RightHand.MetaData1.GetShort("GBAMAX");
```
exectails: `MabiDictionary.Get` returns the default value for the type if the variable doesn't exist. This means if you just get the short, and it doesn't exist, you will get 0, which you can safely get and add to the damage, without having to check if the variable does actually exist.

src/ChannelServer/World/Entities/Creature.cs
```diff
@@ -1867,6 +1867,45 @@ public float GetRndMagicDamage(Skill skill, float baseMin, float baseMax)
...
+					totalMaxDamage += addedBulletDamage;
+				}
+			}
+
+			// Balance
+			var balance = (this.RightHand == null ? this.BalanceBase + this.BalanceBaseMod : this.RightHand.OptionInfo.Balance);
```
exectails: The balance properties already have null checks, you just do `this.BalanceBase + this.BalanceBaseMod + this.RightBalanceMod` and you'll get the correct balance, just like in the other methods that use balance. Easier to read and write.

src/Mabi/Const/Skills.cs
```diff
@@ -530,6 +530,16 @@ public enum SkillId : ushort
 		WirePull = 54201,
 		PuppetsSnare = 54202,
 
+		// Dual Gun
```
exectails: Just out of curiosity, why?

src/ChannelServer/Skills/Guns/BulletStorm.cs
exectails: Exactly.

src/ChannelServer/Skills/Guns/BulletStorm.cs
exectails: Ascending is simply OrderBy^^

src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,337 @@
...
+
+		/// <summary>
+		/// Completes the skill
+		/// </summary>
+		/// <param name="creature"></param>
+		/// <param name="skill"></param>
+		/// <param name="packet"></param>
+		public void Complete(Creature creature, Skill skill, Packet packet)
+		{
+			Send.Effect(creature, 333, (byte)3);
```
exectails: This should probably use the effect id you added.

src/Mabi/Const/Skills.cs
xeroplz: numerical order... it bothered me

src/ChannelServer/Skills/Guns/BulletStorm.cs
xeroplz: Removing the lists has caused issues 
ss 2017-01-03 at 06 37 33

And I really don't understand how creating a list with targetIds and bulletDist values would work in this case :|

src/ChannelServer/Skills/Guns/BulletStorm.cs
xeroplz: would this be a thing for skillhelper?

src/ChannelServer/Skills/Guns/FlashLauncher.cs
xeroplz: Guns use CombatMastery as their target action skill.

src/ChannelServer/Skills/Guns/BulletStorm.cs
xeroplz: The automatic lock doesn't work...

src/ChannelServer/Skills/Guns/FlashLauncher.cs
xeroplz: how would i incorporate that into a void in SkillHelper?

src/ChannelServer/Skills/Guns/BulletStorm.cs
exectails: Because the return type changed, what you get from Take is not the same you get from OrderBy, so you have to cast.

What is it you don't understand about the list?

src/ChannelServer/Skills/Guns/GrappleShot.cs
xeroplz: yes, it works as intended

src/ChannelServer/Skills/Guns/GrappleShot.cs
xeroplz: I have to check if the enemy is targetable in the second cap.Handle because the skill doesn't "hit" friendly entities. Dead enemies aren't targetable so they don't apply to that statement. I have to handle the cap in a second place as a result.

PR #334, opened 2016-06-29
Tir NPC Batch 1

xCryptic:
This is the first batch of updated Tir NPCs. See a complete change list
for all NPCs below:

• added gift weights
• added "who Saved the Goddess" (11001) responses
• added all greetings (ones that were missing)
• added missing adv. keywords
• added missing Phrases in meven
• changed npc name to <npcname/> (except Deian)
• changed all title checks to Title instead of
Player.Titles.SelectedTitle
• changed intros for #241
• updated a couple keywords that had additional checks (like Ferghus's
shop_headman keyword and Duncan's personal_info)
• fixed some keywords being added (some were removed for not being
official, some added since they were missing)
• fixed ModifyRelation values
• removed "base" in class name

### DIFF
--- a/system/scripts/npcs/tir/alissa.cs
+++ b/system/scripts/npcs/tir/alissa.cs
@@ -4,7 +4,7 @@
 // Mill Operator
 //---------------------------------------------------------------------------
 
-public class AlissaBaseScript : NpcScript
+public class AlissaScript : NpcScript
 {
 	const string WindmillPropName = "Uladh_main/town_TirChonaill/windmill_tircho";
 
@@ -21,6 +21,7 @@ public class AlissaBaseScript : NpcScript
 		SetFace(skinColor: 19, eyeType: 10, eyeColor: 148, mouthType: 2);
 		SetStand("human/female/anim/female_natural_stand_npc_alissa");
 		SetLocation(1, 15765, 31015, 120);
+		SetGiftWeights(beauty: 1, individuality: 2, luxury: -1, toughness: 2, utility: 2, rarity: 0, meaning: -1, adult: 2, maniac: -1, anime: 2, sexy: 0);
 
 		EquipItem(Pocket.Face, 3900, 0x00596131, 0x00FFEEC6, 0x006F0017);
 		EquipItem(Pocket.Hair, 3143, 0x00D57527, 0x00D57527, 0x00D57527);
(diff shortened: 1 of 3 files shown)


system/scripts/npcs/tir/alissa.cs
```diff
@@ -243,7 +250,7 @@ public override void Load()
 				break;
 
 			case "skill_windmill":
-				Msg("You can't just operate a windmill.<br/>You need skills. Someone who really knows how. Hehe...<br/>Maybe...someone by the name of Alissa?<br/>If you see someone saying they want to learn the Windmill skill,<br/>bring them to me! I'll show them!");
```
exectails: Out of curiosity, did someone mistakenly use her actual name here instead of the code, or are you replacing all NPC names with the code by default?

exectails:
The commit name could've been better, "Tir NPC Batch 1" is not very descriptive.
ss 2016-06-29 at 09 35 20

system/scripts/npcs/tir/alissa.cs
xCryptic: Huh? Oh, github messed up my description. It should say

changed npc name to `<npcname/>` (except Deian)

I am doing it for all of them.

xCryptic:
Uhh, would Tir NPC Update Batch be better?

system/scripts/npcs/tir/alissa.cs
exectails: Yes, but are you doing that to replace the NPC names, or are you changing the dialog to what it is officially, because whoever did this originally used the names, which was wrong?^^ Or maybe both?

exectails:
Not the PR name, the *commit*, the thing that is displayed in the commit list^^ "Tir NPC Update Batch" would be a little better, though the general convention is to describe what you did with that commit. "Updated XY", "Added YZ", "Fixed Foo", etc.

system/scripts/npcs/tir/alissa.cs
xCryptic: I'm not quite sure what you're getting at. This is the log.

`Msg("You can't just operate a windmill.<br/>You need skills.  Someone who really knows how. Hehe...<br/>Maybe...someone by the name of Alissa?<br/>If you see someone saying they want to learn the Windmill skill,<br/>bring them to me! I'll show them!");`

 But, I was told to replace all NPC names with `<npcname/>` with the exception of Deian.

exectails: You were? Oh^^" Your log answers my question, the original is Alissa, so what we had wasn't incorrect, but you're replacing the names with the code.

PR #336, opened 2016-06-30
Updated Remaining NPCs to the new Greet System

xCryptic:
mymemorysobad

Double checked, all is good.


system/scripts/npcs/bangor/jenifer.cs
```diff
@@ -76,6 +73,32 @@ public override void Load()
...
 
+	private void Greet()
+	{
+		if (Memory <= 0)
+		{
+			Msg(FavorExpression(), L("Welcome to the Bangor Pub. Are you a first-time visitor?"));
+		}
+		else if (Memory == 1)
+		{
+			Msg(FavorExpression(), L("I think I've met you once before... You name is... <username/>, am I right?"));
```
exectails: >You name

Official error?

xCryptic: I would assume so. I didn't always read the messages, just copy & pasted from pale.

PR #346, opened 2016-07-05
Updated Lassar, Malcolm, and Meven

xCryptic:
The next batch of Tir NPC updates. See a complete change list below:

• added gift weights
• added "who Saved the Goddess" (11001) responses
• added missing adv. keywords
• added missing Phrases in meven
• changed npc name to `<npcname/>` (except Deian)
• changed all title checks to Title instead of
Player.Titles.SelectedTitle
• changed intros for #241
• updated a couple keywords that had additional checks (like Ferghus's
shop_headman keyword and Duncan's personal_info)
• fixed some keywords being added (some were removed for not being
official, some added since they were missing)
• fixed ModifyRelation values
• removed "base" in class name

### DIFF
--- a/system/scripts/npcs/tir/lassar.cs
+++ b/system/scripts/npcs/tir/lassar.cs
@@ -1,7 +1,7 @@
 //--- Aura Script -----------------------------------------------------------
-// Lassar, the Magic Instructor in Tir Chonaill
+// Lassar
 //--- Description -----------------------------------------------------------
-// 
+// The Magic Instructor in Tir Chonaill School
 //---------------------------------------------------------------------------
 
 public class LassarScript : NpcScript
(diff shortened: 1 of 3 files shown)


system/scripts/npcs/tir/lassar.cs
```diff
@@ -109,14 +110,18 @@ public override void Load()
...
+							);
 					}
 					else if (result.Points > 1)
 					{
 						if (result.Fails == 0)
 							RndMsg(
 								"I didn't make any mistakes. Hehe",
-								"It has been repaired perfectly!"
+								"It has been repaired perfectly!",
+								"It have been repaired completely. Hehe" // 100% Official http://i.imgur.com/GQH4uOT.png
```
exectails: Not sure if this needs visual proof, are you worried we wouldn't believe you? xD

PR #359, opened 2016-07-16
Added skill "The Fake Spiral Sword"

xeroplz:
- adds skill "The Fake Spiral Sword" from the Fate/stay night [UBW] event.
  Note: optionsets.txt isn't actually changed


src/ChannelServer/Skills/Combat/TheFakeSpiralSword.cs
```diff
@@ -0,0 +1,266 @@
...
+
+		/// <summary>
+		/// Distance the target is knocked back
+		/// </summary>
+		private const int KnockbackDistance = 250;
+
+		/// <summary>
+		/// Effect Enums
+		/// </summary>
+		private enum TheFakeSpiralSwordEffect : byte
```
exectails: Adding an enum for this was the right idea, however, I think it would be better to place it in Mabi.Const.Effects, so it can be used by other classes as well, who knows what we might want to (mis)use certain effects for.

Also, a tip, if you create a static class with consts or readonlys, like our Effect class, you don't need to cast. This is preferable in some situations, when you don't actually use any of the enum features.

src/ChannelServer/Skills/Combat/TheFakeSpiralSword.cs
```diff
@@ -0,0 +1,266 @@
...
+
+			// Check Target
+			if (initTarget == null)
+				return CombatSkillResult.InvalidTarget;
+
+			var attackerPos = attacker.GetPosition();
+			var initTargetPos = initTarget.GetPosition();
+
+			// Check for Collisions
+			if (attacker.Region.Collisions.Any(attacker.GetPosition(), initTargetPos))
```
exectails: You could use attackerPos here and below, instead of getting it again. You could also stop the attacker movement above and use the return value of StopMove, which is the position the creature was stopped at.

src/ChannelServer/Skills/Combat/TheFakeSpiralSword.cs
```diff
@@ -0,0 +1,266 @@
...
+			aAction.Set(AttackerOptions.KnockBackHit1 | AttackerOptions.KnockBackHit2 | AttackerOptions.Result);
+			cap.Add(aAction);
+
+			aAction.Stun = AttackerStun;
+
+			// Get Explosion Radius of Attack
+			var explosionRadius = (int)skill.RankData.Var3 / 2;
+
+			// Get Explosion Targets
+			var targets = attacker.Region.GetCreaturesInRange(initTargetPos, explosionRadius).Where(x => attacker.CanTarget(x) && !attacker.Region.Collisions.Any(initTargetPos, x.GetPosition())).ToList();
```
exectails: Since you don't need the target amount, the ToList call is unnecessary, Where returns an enumerable that you simply use in the foreach below.

You also have problem with the targeting rules here, you're missing checks for NPC on NPC. Check `Creature.GetTargetableCreaturesAround` for an example of what I mean, the method you should probably use for this^^

src/ChannelServer/Skills/Combat/TheFakeSpiralSword.cs
```diff
@@ -0,0 +1,266 @@
...
+			var crit = false;
+			if (attacker.Skills.Has(SkillId.CriticalHit, SkillRank.RF))
+			{
+				var critChance = attacker.GetRightCritChance(0);
+				crit = (rnd.Next(100) < critChance);
+			}
+
+			foreach (var target in targets)
+			{
+				var tAction = new TargetAction(CombatActionType.TakeHit, target, attacker, SkillId.CombatMastery);
```
exectails: I think I've mentioned this before, the last parameter of TargetAction is the attacker skill, passing CombatMastery here and then overwriting it 2 lines below is unnecessary, you can pass `skill.Info.Id` directly.

```
new TargetAction(CombatActionType.TakeHit, target, attacker, skill.Info.Id);
```

PR #364, opened 2016-07-18
Add role playing

DennouNeko:
Core functionality for secondary logins (golems, puppets, RP characters, etc.), switching to Role Playing character and the Three Missing Warriors RP dungeon.

It's a big update and I tried to test as much as I could, but some things may be a bit unstable at first.


Cosrnos:
@exectails @DennouNeko What's the status of this? I'm sure you guys have been talking in chat but can you keep this updated with any info for those of us who prefer to live on github? c:

exectails:
Actually, we haven't talked much yet, and it hasn't been reviewed yet either. My plan is to finish G1 first and then merge this into a new branch, branched off of the G1 branch, to add RP to the RP dungeons.

DennouNeko:
I'll just add, that even if I'd have to rewrite parts of it, overall it's a working solution.
The only thing I'm concerned about is safely ending the RP session, since this may be triggered by, for example killing a mob and there's a chance server will request switching back to main character before finishing sending all combat related packets.

exectails:
Well, generally the RP is during a dungeon or something, where it ends after you've beaten the boss, and maybe watched a cutscene or something. At that point you should normally not be doing anything.

exectails:
Now that I'm looking at this a second time, I want to ask something I ignored two weeks ago: why did you choose to sub-class RP characters from NPCs?

RP characters are player controlled creatures, you're copying lines and lines of PC code to NPC, you're adding conditions to NPC methods because PCs work differently, why not sub-class PlayerCreature, which already has everything you need? Why add `PC || RPNPC` checks, if `PC` would be enough?

DennouNeko:
I wasn't sure back then if I should be using NPC or a PlayerCreature as a base for RPNPC, but from what I remember, the deciding factor was using NPC's ID system to avoid collision with existing NPCs.

exectails:
I see. Them being in the NPC id range is certainly a valid point. I wonder if that's worth the effort and potential for bugs though, when all PC checks need to be changed and code has to be duplicated.

DennouNeko:
Was wondering about that too... In retrospection, probably an ID provider would be better solution than whole NPC class overhaul.
Unfortunately don't think I'll be able to find enough time to do it myself anymore...

exectails:
My first thought was an id provider as well, though I kinda like how the ids come from the respective classes o~o But where exactly they come from is a minor concern I guess. We could also introduce our own RP character id range, that is so far away from the NPCs that a collision is impossible.

exectails:
In the interest of collecting all related information here, as requested by @Cosrnos, I want to recap something we've discussed on the chat a while ago: the way the RP dungeons are being set up.

In this PR setting up the dungeon requires quite a bit of work, setting up the characters, their stats, equip, and skills, etc. manually. What I imagined instead was something like this:

```
[DungeonScript("g1rp_05_tircho_alby_dungeon")]
public class TheThreeMissingWarriorsRPDungeonScript : DungeonScript
{
	public override void OnCreation(Dungeon dungeon)
	{
		dungeon.SetRole(0, "#tarlach");
		dungeon.SetRole(1, "#mari");
		dungeon.SetRole(2, "#ruairi");
	}
```

That would be all the code you need in DungeonScript, with the number specifying who gets the role and the name referencing an actor from data. An initial concern on the chat was that characters appear in different dungeons with different equip/skill/stat configurations, but devCAT "solved" that problem by simply creating copies of the actors as necessary. In the Tarlach Rabbie RP dungeon in G1 for example you're not playing "#tarlach", but "#tarlach2".

What happens manually in this PR, the Dungeon class would handle automatically behind the scenes. Though we might want to add a callback, so it's possible to modify the actors on the fly, if a users wants to change something for some reason.

```
public override void OnActorSetUp(Dungeon dungeon, RpCreature creature)
{
	if (creature.ActorName == "#tarlach")
		creature.Skills.Give(SkillId.Fireball, SkillRank.R1);
}
```

I changed this on my lost copy of this PR, and I'm planning to do it again, unless someone objects or has an even better idea.

exectails:
Closing this PR, now that the feature was added to master. Thanks again, @DennouNeko.

PR #367, opened 2016-07-23
Sitting on props

DennouNeko:
Making the Rest skill search for and use nearby sittable props.

### DIFF
--- a/src/ChannelServer/Skills/Life/Rest.cs
+++ b/src/ChannelServer/Skills/Life/Rest.cs
@@ -199,6 +199,12 @@ namespace Aura.Channel.Skills.Life
 			creature.Temp.SittingProp = null;
 		}
 
+		/// <summary>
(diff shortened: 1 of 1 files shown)


exectails:
Thanks for the contribution, I've been meaning to look into this^^

DennouNeko:
No problem.
I was running around later on, checking the props I know, and seems like later on the range may be increased just a bit. For example, that sofa at Manus' has 3 spots to seat on, but so far got only the middle one to work...

exectails:
Maybe there's more than one variable? I guess there would have to be, because the prop needs to know who sits where, and which spots are taken.

DennouNeko:
Doesn't seem like it... From what I've seen in area file, all 3 spots are added as separate props.

exectails:
Oh. So lazy =<

PR #368, opened 2016-07-24
Sitting on props - improvements

DennouNeko:
Rest checks now slightly wider area, but looks for the closest prop to sit on.

### DIFF
--- a/src/ChannelServer/Skills/Life/Rest.cs
+++ b/src/ChannelServer/Skills/Life/Rest.cs
@@ -41,14 +41,25 @@ namespace Aura.Channel.Skills.Life
 				this.SetUpChair(creature, chairItemEntityId);
 			else
 			{
-				// Find all nearby sittable props and sit on one
+				// Find all nearby sittable props and sit on closest one
+				var crpos = creature.GetPosition();
 				var props = creature.Region.GetProps(prop => prop.HasTag("/sittable/")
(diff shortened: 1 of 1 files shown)


src/ChannelServer/Skills/Life/Rest.cs
```diff
@@ -41,14 +41,25 @@ public override StartStopResult Start(Creature creature, Skill skill, MabiDictio
 				this.SetUpChair(creature, chairItemEntityId);
 			else
 			{
-				// Find all nearby sittable props and sit on one
+				// Find all nearby sittable props and sit on closest one
+				var crpos = creature.GetPosition();
```
exectails: Heh. I was actually thinking of criticizing that you didn't do this before, but then I thought it's not much of a loss, seeing how it will only be checked for sittable props^^

DennouNeko: Yeah... It was the "it's not really necessary with a distance like this" approach ^^'
Now it's a lot easier to sit on certain props, while it still doesn't catch the ones that are way too far.

And now that I think about it, it could be additionally optimized to scan all props and find closest sittable one in one pass, while using squared distance (by not taking the square root of the sum), but not sure if it's really necessary.
Still, GetDistanceSquared could get handy at times when just comparing distances.

exectails: Yea, maybe. Though we primarily use InRange for simple range checks. When I saw your code below I was thinking maybe we should have a method on Entity or somewhere to get the closest other Entity out of a collection.

PR #373, opened 2016-08-02
Updated Tracy

xCryptic:
• Added Gift Weights
• Added Upgrading
• Added Greetings
• Added missing keywords
• Updated `@talk`
• Updated Intro for #241
• Updated about_skill keyword
• Fixed ModifyRelation values
• Other small fixes

### DIFF
--- a/system/scripts/npcs/dugald/tracy.cs
+++ b/system/scripts/npcs/dugald/tracy.cs
@@ -1,7 +1,7 @@
 //--- Aura Script -----------------------------------------------------------
-// Tracy in Ulaid Logging Camp
+// Tracy
 //--- Description -----------------------------------------------------------
-// 
+// The Lumberjack in Ulaid Logging Camp
 //---------------------------------------------------------------------------
 
 public class TracyScript : NpcScript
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/dugald/tracy.cs
```diff
@@ -197,14 +289,15 @@ public override void Load()
...
+
 				switch (await Select())
 				{
 					case "@yes":
 						Msg("Awesome! Repeat after me!<br/>[Tough Guy Tracy!]<br/>Say it 100 times and I'll give you a piece of bread.", Button("Tough Guy Tracy!", "@reply1"), Button("Crazy Dude", "@reply2"));
+
 						switch (await Select())
 						{
 							case "reply1":
-								// NOTE: He asks you to say it 100 times, but I received this message after 1 time
```
xCryptic: This was intentional. No one would click through it 100 times anyway.

exectails: Where is the `Give(bread)` though?^^

xCryptic: I didn't get any bread on NA. Do you think they would just give away free bread? That would be spammable.

exectails: I wasn't serious^^ I don't think you're supposed to get bread, based on what he says.

xCryptic: Oh... welp.

PR #374, opened 2016-08-02
Implement QuestObjectiveCreate

valdeza:
Additional resource: [Sample/Test script (gist)](https://gist.github.com/valdeza/a6ea9e16d87b7c70b5b6300faca5ebcf)
Dependent: All production quest objectives (particularly PTJs asking you to tailor or weave something)

Since events `CreatureCreatedItem` and `CreatureProducedItem` share the quest objective type (same byte ID and packet data structure in NewQuest packets), they share the same event handler (`QuestScript.OnCreatureCreatedOrProducedItem()`).

Additional consideration prior to merging: 
Is [using string comparison](https://github.com/valdeza/aura/blob/5c57b9e02442901abc48017da9c99ccbdc7cb4eb/src/ChannelServer/Scripting/Scripts/QuestScript.cs#L978) to check if skill just used matches the target skill on the quest objective acceptable? 
I was thinking of defining another enum, but seeing it was just a union between `ProductionCategory` and `CreationMethod` it seemed rather redundant.
Comparing `ushort`s would have been possible if `ProductionCategory` had not already been defined as `short`s.

### DIFF
--- a/src/ChannelServer/Scripting/Scripts/QuestScript.cs
+++ b/src/ChannelServer/Scripting/Scripts/QuestScript.cs
@@ -443,6 +443,12 @@ namespace Aura.Channel.Scripting.Scripts
 				ChannelServer.Instance.Events.PlayerRemovesItem += this.OnPlayerReceivesOrRemovesItem;
 			}
 
+			if (objective.Type == ObjectiveType.Create)
+			{
+				ChannelServer.Instance.Events.CreatureCreatedItem -= this.OnCreatureCreatedItem;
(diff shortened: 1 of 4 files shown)


exectails:
Is the objective to actually create the item or is it enough to get it? If it's the latter, it would be simpler to make it a get item objective.

valdeza:
Former. 

In fact, with Malcolm's tailoring and weaving PTJs there are both Create and Collect objectives. Whenever I start the PTJ with the requested items already on hand, I would have the Collect objectives checked off but he still wants me to go outside his shop and weave whatever.

Issue #237, opened 2016-02-16
Part-time Jobs


exectails:
Split from quests, since this will take some more time, and it's important for people to see that we still need a lot of PTJs logged.

Reference: http://wiki.mabinogiworld.com/index.php?title=Part-time_Jobs&oldid=30265

**Part-time jobs**
- [x] **Tir**
- [x] Church
- [x] Grocery Store
- [x] Neighboring
- [x] Healer
- [x] Blacksmith
- [x] General Store
- [x] Inn
- [x] **Dugald Aisle**
- [x] Neighboring
- [x] **Dunbarton**
- [x] Book Store
- [x] Town
- [x] Glenis
- [x] Church
- [x] Healer
- [x] Weapon Shop
- [x] Tailor
- [x] Library
- [x] General Store
- [x] **Bangor**
- [x] Church
- [x] Blacksmith (Elen)
- [x] Mine

DeleteIT:
This is what is required for this particular issue:
http://aura-project.org/forum/topic/173-ptj-dialogues/

exectails:
For those who don't frequently visit the forum, we have a list over there, of what has to be logged for the PTJs to be implemented. Check it out if you'd like to help: http://aura-project.org/forum/topic/173-ptj-dialogues/

gcscaglia:
Bellow are the full logs for Piaras PTJ. These are absolutely *RAW*. There are unrelated packets in the logs, including other PTJs I was doing in between the runs.

[Piaras.zip](https://github.com/aura-project/aura/files/191234/Piaras.zip)

The logs include all possible dialogs (out of order), which are:
* Dialog for when you're already doing a different PTJ.
* Dialog for coming back early without being done.
* Dialog for coming back early after being done.
* Report later dialog.
* Report now and cancel when choosing reward.
* Report now and choose reward.
* Dialog for reporting without having done anything.
* Dialog for reporting with having only done a little.
* Dialog for reporting with having done almost everything.
* Dialog for reporting with having done everything.
* Dialog for PTJ keyword while no PTJs are given out (Before Time).
* Dialog for PTJ keyword while no PTJs are given out (After Time).
* Dialog for declining the PTJ offer.

I'm writing a MabiPale2 plugin to automate logging PTJs and plan on write the actual PTJs afterwards, but I'm posting the logs here in case I end up MIA before that.

Pegwald:
[...]

PR #376, opened 2016-08-06
Add Malcolm PTJ

valdeza:
_~~Warning: This script needs #374 to compile.~~ (Resolved)_

After camping at Malcolm's store for a while, I never got the delivery quest to Trefor so I had to improvise there. Either someone in the future will have to provide it or improve the dialogue I adapted from Ferghus's delivery quest. Other than that, I think everything else has been filled out.

ref: #237 

###### (Oh I'm really not looking forward to Simon's PTJ)


system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -0,0 +1,2103 @@
...
+	/// <summary>
+	/// Depending on the supplied <paramref name="questId"/>, will give the player the needed materials to complete their task.
+	/// </summary>
+	private void GiveStartingPtjItems(int questId, Creature player)
+	{
+		Action<int, int> PlayerGivePattern = (int formId, int useCount) =>
+		{
+			Item pattern = new Item(60600);
+			pattern.MetaData1.SetInt("FORMID", formId);
+			pattern.Data.Durability = useCount * 1000; // Overwrites MaxDurability, else will be clamped to 5000.
```
exectails: Data is the global data, by modifying it you modify the durability of all future items with that item id.

system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -0,0 +1,2103 @@
...
+				npc.Msg(L("Oh, I misunderstood.<br/>I'm sorry."));
+		}
+	}
+
+	/// <summary>
+	/// Depending on the supplied <paramref name="questId"/>, will give the player the needed materials to complete their task.
+	/// </summary>
+	private void GiveStartingPtjItems(int questId, Creature player)
+	{
+		Action<int, int> PlayerGivePattern = (int formId, int useCount) =>
```
exectails: This feels like it should be a Create factory method on Item.

system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -0,0 +1,2103 @@
...
+
+		if (IsEnabled("QuestViewRenewal"))
+			SetCategory(QuestCategory.ById);
+
+		SetType(QuestType.Deliver);
+		SetPtjType(PtjType.GeneralShop);
+		SetLevel(QuestLevel);
+		SetHours(start: 7, report: QuestLevel == QuestLevel.Adv ? 17 : 12, deadline: 19);
+
+		AddObjective("ptj1", L(string.Format("Make 2 {0} (Part-Time Job)", ItemsName)), 0, 0, 0, Create(ItemId, 2, SkillId.Tailoring));
```
exectails: Localization doesn't work like this. The L method (proxy for Localization.Get) looks up the exact string you give it in the localization dictionary, but this item dependent string you're passing here won't be in there. Instead, you'd want this:

`string.Format(L("Make 2 {0} (Part-Time Job)"), L(ItemsName))`

system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -561,7 +552,7 @@ public override void Load()
 			return HookResult.Continue;
 
 		npc.Player.Inventory.Remove(Garment, 1);
-		npc.Notice(L(string.Format("You have given Garment to be Delivered to {0}.", NpcName)));
+		npc.Notice(string.Format(L("You have given Garment to be Delivered to {0}."), L(NpcName)));
```
exectails: `L(NpcName)` won't work either, you always have to wrap the actual string. I guess I should write a guide about this^^

valdeza: Gah. I thought it would be the same way since foreign Mabi (well, or at least Japanese) seems to exchange NPC name in the "gave {the delivery item} to {so-and-so}" string. Or maybe every single occurrence is hardcoded. I'm not sure! 

So for now, I guess I'll mark the times I used `L()` / `Localization.Get()` and hopefully figure out when to separately get localisation for string parametres.

system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -1327,7 +1318,7 @@ public override void Load()
 	{
 		SetId(QuestId);
 		SetName(L("General Shop Part-Time Job"));
-		SetDescription(L(string.Format("This job is tailoring and supplying clothes to the General Shop. Today's order is tailoring [2 {0}], using the materials given for this part-time job. Make sure to bring it to me no earlier than noon. Keep that in mind when delivering the goods, since I can't use them before then.", ItemsName)));
+		SetDescription(string.Format(L("This job is tailoring and supplying clothes to the General Shop. Today's order is tailoring [2 {0}], using the materials given for this part-time job. Make sure to bring it to me no earlier than noon. Keep that in mind when delivering the goods, since I can't use them before then."), L(ItemsName)));
```
valdeza: Marked for review of appropriate `L()` / `Localization.Get()` usage.

system/scripts/quests/ptj/general_malcolm.cs
valdeza: After what was read [here](https://github.com/aura-project/aura/pull/383#discussion_r75103854), would it actually be the following?
```csharp
// ItemName is the singular forme. Currently not defined in script.
string.Format(L("Make 2 {0} (Part-Time Job)"), LN(ItemName, ItemsName, 2))
```

system/scripts/quests/ptj/general_malcolm.cs
valdeza: Confirm: correct usage?
```csharp
SetDescription(string.Format(L("This job // ... // is tailoring [2 {0}], using the materials // ... // before then."), LN(ItemName, ItemsName, 2)));
```

system/scripts/quests/ptj/general_malcolm.cs
exectails: No, because the amount is static, it's not a plural localization, you just localize the item names in this script, with the knowledge that it will always be two. I've just noticed that you're using properties for the item names though, you have to wrap *those* in L, not those getter calls.

 I guess for localization it would be easier if you had the entire string though... for context.

Localization like this is most definitely a hassle, but I like the idea of us providing a way to *properly* translate everything that is sent to the client. Especially since Mabi has never been released in the west in any language but English.

valdeza: All right, then I'll make the objective descriptions into simple, non-parameterised strings. 

But, hypothetically, if we were to stick to what we have now, why wrap the getter return values in L when [`GetPluralString(...)`](https://github.com/exectails/nettext/blob/master/nettext/PoFile.cs#L366) is expecting string IDs and not post-translation strings?

My interpretation of what you said:
```csharp
public abstract int N { get; }
public string TranslatedString { get { return L("untranslated string"); } }
public string TranslatedStrings { get { return L("untranslated strings"); } }
public string Foo() { return LN(TranslatedString, TranslatedStrings, N; }
```

I mean, I _think_ it would work anyway since `GetPluralString(...)` would not find the key and return the translated strings, but we're searching the `_storage` dictionary twice with no meaningful work on the second search.
_Edit: It might be grammatically incorrect for languages with multiple plural formes. Linguistics is hard .-._

exectails: Your interpretation is correct if you have a string that actually has plural forms.

valdeza: Um. Sorry. I know you already suggested 
`string.Format(L("Make 2 {0} (Part-Time Job)"), L(ItemsName))` , but
> I guess for localization it would be easier if you had the entire string though... for context.

wouldn't this statement be suggesting to go back to 
`L(string.Format("Make 2 {0} (Part-Time Job)", ItemsName))` 
when the "item dependent string [being passed] here won't be in there"? The latter should essentially be the same as 
`L("Make 2 Foobar's Clothes (Part-Time Job)")` , right? 
Unless `String.Format(...)` and string-type getter return values do something to strings that will affect retrieval from the `_storage` dictionary that I'm not aware of.

Just wanted to be perfectly clear before I end up pushing, like, five trivial commits tweaking Localization method calls on this and the other PTJ scripts.

exectails: No, because then you'd be passing a string that couldn't possibly be in the localization dictionary to L again.

valdeza: Ohh. So Poedit is explicitly searching for strings in Localization method calls within source code files.

* `string.Format(L("Make 2 {0} (Part-Time Job)"), L(ItemsName))`  
returns "Make 2 {0} (Part-Time Job)"
* `L(string.Format("Make 2 {0} (Part-Time Job)", ItemsName))`  
returns _nothing_.
* `L("Make 2 Foobar's Clothes (Part-Time Job)")`  
returns "Make 2 Foobar's Clothes (Part-Time Job)"

And you told me to wrap the string return values with L so that it would be picked up by Poedit, since Poedit will ignore `L(PropertyName)` (apparently searching for quotation marks). But using your suggestion, the translator will lose context, as they see strings "Make 2 {0} (Part-Time Job)" and "Foobar's Clothes" separately. So to facilitate more accurate translations, you are saying the third bullet point is preferred.

Am I understanding the Localization systemme correctly?

exectails: Almost^^ The second point returns something as well: whatever you put in.

```csharp
var itemsName = "Foobar's Clothes";
var phrase = string.Format("Make 2 {0} (Part-Time Job)", ItemsName); // Make Foobar's Clothes (Part-Time Job)
var translated = L(phrase); // Make Foobar's Clothes (Part-Time Job)
```

Since that phrase won't exist in the dictionary for every single item it will simply return the default, which is what came in. Unless somewhere there's a `L("Make Foobar's Clothes (Part-Time Job)")` to be found, which the parser would pick up and add to the dictionary as a translatable string.

valdeza: Oh whenever I said 'return', I meant when having Poedit parse through the source code file. 
`L(string.Format("Make 2 {0} (Part-Time Job)", ItemsName))` does not generate any "Source text".

exectails: Ah, yes, that's correct.

Issue #343, opened 2016-07-04
Unable to specify additional conditions on Gather or Collect quest objectives


valdeza:
Current implementations ([QuestObjectiveGather](https://github.com/aura-project/aura/blob/159840e538a74866639f83d793cff4c63064fd23/src/ChannelServer/World/Quests/Objectives.cs#L225), [QuestObjectiveCollect](https://github.com/aura-project/aura/blob/159840e538a74866639f83d793cff4c63064fd23/src/ChannelServer/World/Quests/Objectives.cs#L74)) only allow scripts to specify an item and an amount. Some future quest objectives will specify additional conditions on the requested item, such as cooking food of at least a certain quality or fishing for carp of at least a certain size.

dependent: pending link to unimplemented npc:caitin.ptj: carp fishing

valdeza:
### Proposal_
**Edit: This proposal has a problem. See "Possibly affected files" below.**

Provide the following method overloads:
```csharp
// aura/src/ChannelServer/Scripting/Scripts/QuestScript.cs
protected QuestObjective QuestScript.Collect(int itemId, int amount, Func<Item, bool> conditions);
protected QuestObjective QuestScript.Gather(int itemId, int amount, Func<Item, bool> conditions);
```

#### Possibly affected files_
* _aura/src/ChannelServer/World/Quests/Objectives.cs_  
Will need an overloaded constructor to use `conditions`. MetaData will probably need to be set with the required conditions (unconfirmed, can't choose which PTJ I get).
* _aura/src/ChannelServer/Scripting/Scripts/QuestScript.cs_  
Methods attending to `ObjectiveType.Gather` and `ObjectiveType.Collect` will need to be updated to query item metadata.  
**Problem: Only item IDs are used throughout this code file. Need to somehow access metadata.**

#### Usage example_
```csharp
AddObjective("ptj", L("Deliver a 10cm+ Brifne Carp"), 0, 0, 0, 
	Collect(MISSINGNO227, 1, (item) => {return item.MetaData1.GetFloat("SIZE") > 10.0f;}));
// "SIZE" might not be the key name, but is used for illustrative purposes
```

#### Considerations_
Those wishing to make custom quests should expect a little bit more coding complexity in exchange for more complex quest conditions.  
Or should this be abstracted away somehow so that the scripter does not need to know lambda expressions and to call `Item.MetaData1.Get<T>()`?

exectails:
[...]

PR #377, opened 2016-08-07
Bug fix: Do not count WIP items for Collect quest objectives

valdeza:
Collect quest objectives will no longer count production skill works in progress (e.g. half-finished tailored Cores' Healers Dress).

This was solved by adding checks for whether PRGRATE metadata exists on the item to be counted. The modified item counting functions now have an optional bool parametre to specify whether or not to perform PRGRATE checks (made optional to avoid needing to change all count function call sites).

### DIFF
--- a/src/ChannelServer/Scripting/Scripts/QuestScript.cs
+++ b/src/ChannelServer/Scripting/Scripts/QuestScript.cs
@@ -712,7 +712,7 @@ namespace Aura.Channel.Scripting.Scripts
 
 					case ObjectiveType.Collect:
 						var itemId = (objective as QuestObjectiveCollect).ItemId;
-						var count = creature.Inventory.Count(itemId);
+						var count = creature.Inventory.Count(itemId, false);
 
 						if (!progress.Done && count >= objective.Amount)
 							quest.SetDone(progress.Ident);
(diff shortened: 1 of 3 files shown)


exectails:
Thanks for your contribution! I wonder if we can do better here though.

Your implementation is very specific, with a bool parameter that does *one specific* thing, in classes that should in my opinion generally don't have to worry about such specific things. Imagine next week we get another situation like this, where certain items in certain states aren't counted, we would have to add another bool, and another, and another.

As an alternative solution I'd propose adding an overload for Count that takes just a lambda, so you can specify which items should be counted yourself, without the inventory and pocket classes needing any specific information. This way we also don't break any existing code.

Quicky example:
```csharp
public int Count(Func<Item, bool> predicate)
{
	lock (_pockets)
		return _pockets.Values.Where(a => !InvisiblePockets.Contains(a.Pocket))
			.Sum(pocket => pocket.CountItem(predicate));
}

public override int CountItem(Func<Item, bool> predicate)
{
	return _items.Values.Where(predicate)
		.Aggregate(0, (current, item) => current + item.Info.Amount);
}
```

What do you think?

valdeza:
I think it [sounds familiar](https://github.com/aura-project/aura/issues/343) haha. Also, it's always better to only have to change something from [one place](link_pending) rather than three.

exectails:
Certainly familiar, but fundamentally different^^

Yes, this is much better, thanks!

PR #380, opened 2016-08-13
Abstract ptj script

marlesu:
Pulled out common code in ptj scripts and put them in an abstract class.


src/ChannelServer/Scripting/Scripts/AbstractPartTimeJobScript.cs
```diff
@@ -0,0 +1,226 @@
...
+		public abstract int[] QuestIds { get; }
+
+		// The npc name id that is taken as a key to hook this script to ptj keyword. eg. _piaras
+		public abstract string NpcNameId { get; }
+		// The name of the part time job that is shown to the user. eg. Piaras's Inn Part-time Job
+		public abstract string PtjName { get; }
+		// The description of the part time job that is shown to the user. eg. Looking for help with delivering goods to Inn.
+		public abstract string PtjDescription { get; }
+
+		const string HOOK_NAME_AFTER_INTRO = "after_intro";
```
exectails: Constants use PascalCase, as per our coding conventions. You can find them in the [readme](https://github.com/aura-project/aura#coding-conventions).

src/ChannelServer/Scripting/Scripts/AbstractPartTimeJobScript.cs
```diff
@@ -0,0 +1,226 @@
...
+
+		// Must set all of these strings in the struct
+		struct Dialogs
+		{
+			public readonly string doingPtjForAnotherNpc;
+
+			public readonly string earlyReport_finished;
+			public readonly string earlyReport_notFinished;
+
+			public readonly string report_dialogText;
```
exectails: Constants use PascalCase, as per our coding conventions. You can find them in the [readme](https://github.com/aura-project/aura#coding-conventions).

src/ChannelServer/Scripting/Scripts/AbstractPartTimeJobScript.cs
```diff
@@ -0,0 +1,226 @@
...
+			// Check if PTJ is in progress
+			if (npc.DoingPtjForNpc())
+			{
+				var result = npc.GetPtjResult();
+
+				// Check if report time
+				if (!npc.ErinnHour(Report, Deadline))
+				{
+					if (result == QuestResult.Perfect)
+						npc.Msg(L(PtjDialogs.earlyReport_finished));
```
exectails: `L` is short for `Localization.Get`, which retrieves the passed string from the localization dictionary, and the tool that builds the localization dictionary looks up those strings by searching for the first string parameter on all calls to `L` or `Localization.Get`. You either need to pass a string directly, or you wrap the strings in `L` wherever they're set and you remove the calls here.

src/LoginServer/Database/LoginDb.cs
```diff
@@ -326,7 +326,7 @@ public List<Item> GetEquipment(long creatureId)
 						var item = new Item();
 						item.Id = reader.GetInt64("entityId");
 						item.Info.Id = reader.GetInt32("itemId");
-						item.Info.Pocket = (Pocket)reader.GetByte("pocket");
+                        item.Info.Pocket = (Pocket)reader.GetInt32("pocket");
```
exectails: Please set your IDE to use tabs, as per our coding conventions. You can find them in the [readme](https://github.com/aura-project/aura#coding-conventions).

exectails:
Thanks for your contribution, but as is, this class might not be very useful.

There are 2 reasons there's no abstraction for PTJs (yet?), the first is that we don't know all the requirements yet. For example, some have more responses than others to certain situations, others need additional, or less hooks, and maybe the next one we add requires a completely different structure.

While all that can technically be abstracted, it will potentially require a constant rewriting of all PTJs, to factor in changes, and the more abstraction we add, the more complicated the scripts become, to a point where only coders would be able to follow them.

This brings me to the second reason, abstraction complicates things. Scripts are ideally not only dynamic, but also simple enough for non coders to follow and make simple changes, if feasible. Add a GiveItem here, change a message there, this becomes much more complicated if this requires an understanding of one or multiple underlying abstractions, overloads, overrides, etc.

I agree that the PTJ scripts might be a little messy at the moment, but in my opinion it might be a little early to abstract them in this way. What do you think?

marlesu:
Thanks for reviewing my pull request. I'm sorry that I didn't correctly follow all the naming conventions. I'm still transitioning from Java naming convention to C#.

I will go ahead and fix all those names.

marlesu:
Hi, 
the reason I thought abstraction could be an improvement was because it would mean the non-coders could simply fill out the ptj information and just write the ptj classes to include the rewards/requirements.

On top of that, vast majority of npcs have identical conversation trees to ptj. The ones with unique conversation tree could simply override the AboutArbeit method.

As for future hooks needing to be added/removed, I thought it was easier to just change the base class than to go to each individual copy-pasted ptj scripts and change them.

I understand if this is too early for abstraction (although I think they could never be too early). We could just keep this branch on hold.

exectails:
>As for future hooks needing to be added/removed

What I meant with the hooks is that *some* NPCs need more than others.

>I think they could never be too early

Normally you'd be right of course, but normally you know your requirements. With server emulation you either have to really do the entire research first, or you won't know your requirements until your done. There are systems in Aura that have been rewritten half a dozen times over the years, because we kept finding new requirements, and if requirements changed with the PTJs, you potentially have to change all of them again and again. I've become careful with early abstraction, because it's always me who had to clean up the "mess" =<

marlesu:
okay I'll close this for now.

PR #381, opened 2016-08-15
Add CreatePattern factory method on Item.cs

valdeza:
(Also applies to blacksmith manuals.)

Sample usage:
```csharp
// Apprentice Sewing Pattern - Mongo Traveler Suit (F), 40 uses
npc.Player.GiveItem(Item.CreatePattern(60600, 10107, 40));
```

re: https://github.com/aura-project/aura/pull/376/files#r74115977 
Is this what you meant, @exectails ?

> Data is the global data, by modifying it you modify the durability of all future items with that item id.

Durability is now modified through `Item.OptionInfo` . When I spawn additional patterns, default durability is untouched.

### DIFF
--- a/src/ChannelServer/World/Entities/Item.cs
+++ b/src/ChannelServer/World/Entities/Item.cs
@@ -784,6 +784,27 @@ namespace Aura.Channel.World.Entities
 		}
 
 		/// <summary>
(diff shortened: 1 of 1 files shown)


exectails:
It is :+1:

PR #383, opened 2016-08-16
Add Tracy PTJ, Big Order of Firewood collection quest

valdeza:
_Edit: Missing: Dialogue for a first-time worker accepting and rejecting Tracy's PTJ._

Big Order of Firewood is a possible reward for turning in an advanced level PTJ. However, quest details have been improvised since never in my life will I be eligible to do advanced PTJs.

Also, Piaras had incorrect quest IDs ([mabipcap](https://github.com/valdeza/aura/blob/resrch_ptj-dialogue/etc/ptj-dialogue/tir-chonaill/piaras%20ferghus%20malcolm%20%40gcscaglia.mabipcap.raw.txt)), causing quest ID conflicts. Fixed!

ref: #237

### DIFF
--- a/system/scripts/quests/ptj/general_tracy.cs
+++ b/system/scripts/quests/ptj/general_tracy.cs
@@ -184,18 +184,18 @@ public class TracyPtjScript : GeneralScript
 		if (await npc.Select() == "@accept")
 		{
 			if (npc.GetPtjDoneCount(JobType) == 0)
-				msg = L("(missing): first time accepting PTJ offer");
+				npc.Msg(L("(missing): first time accepting PTJ offer"));
 			else
-				msg = L("Outstanding! Go on, man.");
+				npc.Msg(L("Outstanding! Go on, man."));
 
 			npc.StartPtj(randomPtj);
 		}
 		else
 		{
 			if (npc.GetPtjDoneCount(JobType) == 0)
-				msg = L("(missing): first time declining PTJ offer");
+				npc.Msg(L("(missing): first time declining PTJ offer"));
 			else
-				msg = L("Well... if you're not interested...");
+				npc.Msg(L("Well... if you're not interested..."));
 		}
 	}
 }
--- a/system/scripts/quests/ptj/general_tracy.cs
+++ b/system/scripts/quests/ptj/general_tracy.cs
@@ -1,15 +1,3 @@
-#region TEMP_NOTES
-// If you see this region in a pull request, 
-// **I contain undefined values and should not be merged into master!**
-
-// Class name resolver - go go gadget intellisense
-using Aura.Mabi;
-using Aura.Mabi.Const;
-using System.Threading.Tasks;
-using Aura.Channel.Scripting.Scripts;
-// Temporarily place this file under ChannelServer.csproj to resolve rest of dependencies
-#endregion
-
 //--- Aura Script -----------------------------------------------------------
 // Tracy's Firewood-Chopping (General) Part-Time Job
 //--- Description -----------------------------------------------------------
@@ -29,7 +17,9 @@ public class TracyPtjScript : GeneralScript
 
 	readonly int[] QuestIds = new int[]
 	{
-		?
+		500102, // Basic  Gather 10 Pieces of Firewood
+		500132, // Int    Gather 20 Pieces of Firewood
+		500162, // Adv    Gather 30 Pieces of Firewood
 	};
 
 	public override void Load()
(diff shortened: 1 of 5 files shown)


system/scripts/quests/ptj/general_tracy.cs
```diff
@@ -0,0 +1,300 @@
...
+	protected abstract int QuestId { get; }
+	protected abstract int ItemCount { get; }
+	protected abstract QuestLevel QuestLevel { get; }
+	protected abstract void AddRewards();
+
+	public override void Load()
+	{
+		SetId(QuestId);
+		SetName(L("Firewood-Chopping PTJ"));
+		SetDescription(string.Format(L("A part-time job to chop firewood from the trees in the forest. Collect [{0} pieces of firewood]. First, you need an axe to chop firewood."), L(ItemCount.ToString())));
```
valdeza: Marked for review of appropriate `L()` / `Localization.Get()` usage.

system/scripts/quests/ptj/general_tracy.cs
```diff
@@ -0,0 +1,300 @@
...
+
+		if (IsEnabled("QuestViewRenewal"))
+			SetCategory(QuestCategory.ById);
+
+		SetType(QuestType.Deliver);
+		SetPtjType(PtjType.General);
+		SetLevel(QuestLevel);
+		SetHours(start: 7, report: 9, deadline: 19);
+
+		AddObjective("ptj", string.Format(L("Collect {0} Pieces of Firewood"), L(ItemCount.ToString())), 0, 0, 0, Collect(63002, ItemCount));
```
valdeza: Marked for review of appropriate `L()` / `Localization.Get()` usage.

system/scripts/quests/ptj/general_tracy.cs
exectails: We really need that guide^^ `L`/`Localization.Get` looks up the *string* you pass it, in the localization dictionary. For example, `L("Hello")` would look for translations for the string "Hello". Like if someone has a German dictionary loaded, `L` would return "Hallo". If no localization dictionary is loaded, or the given string wasn't translated, the string is returned back, as the default.

Passing an int to L is pointless, as there won't be an entry for it.

There's another gotcha here though, which is plurals: "[{0} pieces of firewood]". The phrase depends on ItemCount. Even if it's always more than one you need to use plural localization, because there are many languages that have more than 2 plural forms like in English.

```csharp
// simple string to string
L("[1 piece of firewood]");

// Plural localization, passing singular and plural form to LN (Localization.GetPlural),
// and a number that's used to determine which to use.
// Afterwards format to get the number into the string.
string.Format( LN("[{0} piece of firewood]", "[{0} pieces of firewood]", ItemCount), ItemCount );
```

Until we have that guide, you can read a *little* more about it here:
http://aura-project.org/forum/topic/187-project-log-2016-02-15
https://github.com/aura-project/aura/blob/master/system/localization/readme.txt

valdeza: So the correct usage here would be the following?
```csharp
SetDescription(string.Format(LN(
	"A part-time job to chop firewood from the trees in the forest. Collect [{0} piece of firewood]. First, you need an axe to chop firewood.", 
	"A part-time job to chop firewood from the trees in the forest. Collect [{0} pieces of firewood]. First, you need an axe to chop firewood.",
	ItemCount), ItemCount));
```

And when would one use 
`GetParticular(string context, string id)` or 
`GetParticularPlural(string context, string id, string id_plural, int n)`?
Is `context` a unique string ID to provide an alternate translation in a situation where the one provided by `Get()` / `GetPlural()` would be undesirable?
What would be an example of one such situation?

system/scripts/quests/ptj/general_tracy.cs
valdeza: Confirm: correct usage?
```csharp
AddObjective("ptj", string.Format(LN(
	"Collect {0} Piece of Firewood", 
	"Collect {0} Pieces of Firewood", 
	ItemCount), ItemCount), 
	0, 0, 0, Collect(63002, ItemCount));
```

exectails: Yup.

system/scripts/quests/ptj/general_tracy.cs
exectails: One example would be the word "file", which might be a computer file or maybe a record of sorts, depending on the *context*. If you're using both meanings, you need a way to differentiate them, so they can be translated properly. For example, in German the two above would be "Datei" and "Akte" respectively, there's not one word for both.

system/scripts/quests/ptj/general_tracy.cs
```diff
@@ -0,0 +1,303 @@
...
+			return;
+		}
+
+		// Offer PTJ
+		var randomPtj = npc.RandomPtj(JobType, QuestIds);
+		var msg = "";
+
+		if (npc.GetPtjDoneCount(JobType) == 0)
+		{
+			npc.Msg(L("Good. You want a logging job?<br/>I was actually a little bored working alone.<br/>I can use some help. If you're good enough, I can pay you more."));
```
exectails: Just a note, you could simplify this a little by putting both messages into msg with '<p/>', which basically clears the dialog. This is also another thing that makes localization easier, because the person translating gets both messages in one string, to get the full meaning.

```csharp
msg = L("Good. You want a logging job? [...] If you're good enough, I can pay you more.<p/>Want to give it a try?");
```

PR #386, opened 2016-08-18
Add Eavan PTJ

valdeza:
Taking a PTJ from Eavan is now possible but starting at intermediate-level jobs, there is a chance for you to run into the following:

Missing dialogue:
* first time worker PTJ inquiry
* first time accepting PTJ offer
* first time declining PTJ offer
* 1-star PTJ turn in response
* 3-star PTJ turn in response

Improvised dialogue:
* Item delivery: Tracy
* Extended item delivery: Manus -> Austeyn
* Extended item delivery: Nerys -> Simon

ref: #237

### DIFF
--- a/system/scripts/quests/ptj/adventurerassn_eavan.cs
+++ b/system/scripts/quests/ptj/adventurerassn_eavan.cs
@@ -1,17 +1,5 @@
-#region TEMP_NOTES
-// If you see this region in a pull request, 
-// **I contain undefined values and should not be merged into master!**
-
-// Class name resolver - go go gadget intellisense
-using Aura.Mabi;
-using Aura.Mabi.Const;
-using System.Threading.Tasks;
-using Aura.Channel.Scripting.Scripts;
-// Temporarily place this file under ChannelServer.csproj to resolve rest of dependencies
-#endregion
-
 //--- Aura Script -----------------------------------------------------------
-// Eavan's Adven. Assoc. Part-Time Job
+// Eavan's Adventurer's Association Part-Time Job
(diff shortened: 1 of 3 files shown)


system/scripts/quests/ptj/adventurerassn_eavan.cs
```diff
@@ -0,0 +1,984 @@
...
+				npc.CompletePtj(reply);
+				remaining--;
+
+				// Result msg
+				if (result == QuestResult.Perfect)
+				{
+					npc.Msg(npc.FavorExpression(), L("Good. Just as I asked you to do.<br/>Thank you very much."));
+					npc.ModifyRelation(0, Random(3), 0);
+				}
+				else if (result == QuestResult.Mid)
```
exectails: She doesn't actually have jobs that can end up in mid and low results, does she? I guess we don't need those cases then, missing dialog messages only make people think we're missing something^^

valdeza: Well, if someone could get to at least her intermediate-level jobs, there could be a chance they either get a mid or low level result by only doing the first part. Might depend on how RNG decides to round the floating point.

exectails: The question I ask myself is whether you can get anything but Perfect or None on a PTJ that doesn't have a collect objective. Actually, in Aura you currently can't, because the quest result will be None if you didn't complete the last objective.

>Well, if someone could get to at least her intermediate-level jobs

Somehow I doubt anybody has ever done 50 PTJs for Eeavan^^

PR #389, opened 2016-08-20
Add Aeira's PTJ

valdeza:
Taking a PTJ from Aeira is now possible but starting at intermediate-level jobs, there is a chance to run into the following:

Missing dialogue:
* 1-star PTJ turn in response
* 3-star PTJ turn in response

Improvised dialogue:
* Extended item delivery: Nerys -> Stewart
* Extended item delivery: Manus -> Walter


exectails:
Keep up the great work! =)

PR #391, opened 2016-08-23
Add Glenis PTJ, Big Order of Milk collection quest

valdeza:
### Important Note_
Both this and Caitin's PTJ scripts share quest IDs for berry gathering (/5011[036]1/), so this script must load last. If, in the future, script loading were to become multithreaded, this might be a problem.

### Issues_
Starting at intermediate-level jobs, there is a chance to run into improvised dialogue for the following PTJs:
* Extended item delivery: Aeira -> Stewart
* Extended item delivery; Aranwen -> Stewart

---

### Resolved_
* First-time worker dialogue
* Cows can now be milked
* Able to provide player with starting items for Big Order of Milk. Quest reverted to original.

ref: #237


system/scripts/quests/ptj/grocery_glenis.cs
```diff
@@ -0,0 +1,862 @@
...
+	/// <summary>
+	/// Depending on the supplied <paramref name="questId"/>, will give <param name="player"/> the needed materials to complete their task.
+	/// </summary>
+	private void GiveStartingPtjItems(int questId, Creature player)
+	{
+		const int EmptyBottle_PTJ = 63022;
+
+		switch (questId)
+		{
+			case 501102: // Basic  Gather  5 Milk Bottles
```
exectails: Maybe we need a `OnReceive` callback in QuestScript to handle giving items like these. That would also solve the problem you described with the other, non-PTJ quest.

exectails:
I've added the receive event, that should come in handy. Especially since you already had to work around that restriction in other PTJs iirc =|

I also got the missing dialog:
```csharp
// inquiry
npc.Msg(L("Are you looking for work?<br/>You don't seem to have any experience in this line of work. Are you sure you can handle it?<br/>Well, why don't you get started on this now?<br/>Any type of work is difficult at first, but you get used to it as you gain more experience, you see.<p/>Well, how about it?"));

// reject
npc.Msg(L("Ha ha. You don't have to get so intimidated.<br/>All right. Come by next time."));

// accept
npc.Msg(L("Yes, yes. Good idea.<br/>Now, I'll see you by the deadline.<br/>Even if you don't finish everything, at least come file a report."));
```

And I checked milking and fixed it. I guess we added a tool durability check some time, and since bottles don't have a durability, they were seen as a broken tool.

valdeza:
Cows can be milked, 
Adelia stream can be gathered from, 
and OnReceive() works just fine!

PR #392, opened 2016-08-24
quickbugfix: Preserve PTJ bottles on consumption

valdeza:
Currently, consuming _Milk (PTJ) (ID: 63023)_ 
leaves you with an _Empty Bottle (ID: 63020)_ 
instead of an _Empty Bottle (PTJ) (ID: 63022)_.
This text _happens to be shaped a rectangle._


exectails:
Thanks for the fix. This made me realize we might want a more generalized solution. Even if there are currently only 4 items that need a replacement on use, more could easily be added by devCAT, and maybe users have different ideas for replacements as well.

I've just added a field for this to the item db, so I'm going to close this PR.

PR #399, opened 2016-08-29
Fix for unimplemented Dressing Room freezing client.

DennouNeko:
Improved sending the "Failed to open the Dressing Room" packet.
It's based on actual denial from officials.

### DIFF
--- a/src/ChannelServer/Network/Handlers/Misc.cs
+++ b/src/ChannelServer/Network/Handlers/Misc.cs
@@ -528,8 +528,7 @@ namespace Aura.Channel.Network.Handlers
 		{
 			var creature = client.GetCreatureSafe(packet.Id);
 
-			Send.MsgBox(creature, Localization.Get("This feature hasn't been implemented yet."));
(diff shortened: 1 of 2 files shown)


src/ChannelServer/Network/Handlers/Misc.cs
```diff
@@ -543,7 +542,8 @@ public void DressingRoomClose(ChannelClient client, Packet packet)
 		{
 			var creature = client.GetCreatureSafe(packet.Id);
 
-			Send.DressingRoomCloseR(creature, true);
+			Send.DressingRoomCloseR(creature, false);
```
exectails: A negative response to close? Huh, weird. Well, if that's official and works o.o

DennouNeko: It may be an indicator, if data were saved. Doubt they'd want to prevent client from /closing/ the window.
Anyway, that byte is for sure zero in case of denial.

PR #401, opened 2016-09-01
Simultaneous Attacks (and Dependencies from old xKamuna/aura repository)

xKamuna:
Dependencies include:
Updates to combat-related code.
New Creature variables.
New Math2 and Creature functions.

See pull request code for specific changes.

### DIFF
--- a/src/ChannelServer/Network/Handlers/Combat.cs
+++ b/src/ChannelServer/Network/Handlers/Combat.cs
@@ -90,6 +90,15 @@ namespace Aura.Channel.Network.Handlers
 
 			// Purpose unknown, without this the client doesn't seem to
 			// accept the stun time, you can spam attacks.
+			// This updates the target so that it knows that you're attacking!  It lets you determine when someone is trying to attack. - Kamuna
+			if (targetEntityId == 0)
+			{
+				creature.AttemptingAttack = false;
+			}
+			else
+			{
+				creature.AttemptingAttack = true;
+			}
 			Send.CombatTargetUpdate(creature, targetEntityId);
 		}
 
@@ -121,6 +130,27 @@ namespace Aura.Channel.Network.Handlers
 
 			var creature = client.GetCreatureSafe(packet.Id);
 
+			// Get skill
+			var skill = creature.Skills.ActiveSkill;
+			var combatMastery = creature.Skills.Get(SkillId.CombatMastery);
+			if (skill == null && (skill = combatMastery) == null)
+			{
+				Log.Warning("CombatAttack: Creature '{0}' doesn't have Combat Mastery.", creature.Name);
+				goto L_End;
+			}
+			if (skill == combatMastery || skill == null)
+			{
+				if (targetEntityId == 0)
+				{
+					creature.AttemptingAttack = false;
+				}
+				else
+				{
+					creature.AttemptingAttack = true;
+				}
+				Send.CombatTargetUpdate(creature, targetEntityId);
+			}
+
 			// Check lock
 			if (!creature.Can(Locks.Attack))
 			{
@@ -136,14 +166,6 @@ namespace Aura.Channel.Network.Handlers
 			// Check Stun
 			if (creature.IsStunned)
 				goto L_End;
-
-			// Get skill
-			var skill = creature.Skills.ActiveSkill;
-			if (skill == null && (skill = creature.Skills.Get(SkillId.CombatMastery)) == null)
-			{
-				Log.Warning("CombatAttack: Creature '{0}' doesn't have Combat Mastery.", creature.Name);
-				goto L_End;
-			}
 
 			// Get handler
 			var skillHandler = ChannelServer.Instance.SkillManager.GetHandler<ISkillHandler>(skill.Info.Id);
--- a/src/ChannelServer/Scripting/Scripts/AiScript.cs
+++ b/src/ChannelServer/Scripting/Scripts/AiScript.cs
@@ -242,6 +242,11 @@ namespace Aura.Channel.Scripting.Scripts
 
 				this.SelectState();
 
+				if (this.Creature.IsKnockedDown || this.Creature.WasKnockedBack)
+				{
+					this.Creature.AttemptingAttack = true;
+				}
+
 				// Stop and clear if stunned
 				if (this.Creature.IsStunned)
 				{
@@ -269,7 +274,7 @@ namespace Aura.Channel.Scripting.Scripts
 							case AiState.Aggro: this.SwitchAction(Aggro); break;
 							case AiState.Love: this.SwitchAction(Love); break;
 						}
-
+						this.Creature.AttemptingAttack = false;
 						_curAction.MoveNext();
 					}
 				}
@@ -304,7 +309,10 @@ namespace Aura.Channel.Scripting.Scripts
 			_state = AiState.Idle;
 
 			if (this.Creature.IsInBattleStance)
+			{
 				this.Creature.IsInBattleStance = false;
+				this.Creature.AttemptingAttack = false;
+			}
 
 			if (this.Creature.Target != null)
 			{
@@ -1352,10 +1360,19 @@ namespace Aura.Channel.Scripting.Scripts
 					yield break;
 
 				// Attack
+				if (!this.Creature.AttemptingAttack)
+				{
+					this.Creature.AttemptingAttack = true;
+				}
+				if (!this.Creature.CanAttack(this.Creature.Target))
+				{
+					yield return true;
+				}
 				var result = skillHandler.Use(this.Creature, skill, this.Creature.Target.EntityId);
 				if (result == CombatSkillResult.Okay)
 				{
 					// Stop when max attack count is reached
+					this.Creature.AttemptingAttack = false;
 					if (++i >= count)
 						break;
 
--- a/src/ChannelServer/Skills/Combat/CombatMastery.cs
+++ b/src/ChannelServer/Skills/Combat/CombatMastery.cs
@@ -49,17 +49,122 @@ namespace Aura.Channel.Skills.Combat
 		/// <returns></returns>
 		public CombatSkillResult Use(Creature attacker, Skill skill, long targetEntityId)
 		{
-			if (attacker.IsStunned)
-				return CombatSkillResult.Okay;
-
 			var mainTarget = attacker.Region.GetCreature(targetEntityId);
 			if (mainTarget == null)
 				return CombatSkillResult.Okay;
 
-			if (!attacker.GetPosition().InRange(mainTarget.GetPosition(), attacker.AttackRangeFor(mainTarget)))
-				return CombatSkillResult.OutOfRange;
+			if (mainTarget.IsNotReadyToBeHit && attacker.InterceptingSkillId == SkillId.None)
+				return CombatSkillResult.Okay;
+
+			if ((attacker.IsStunned || attacker.IsOnAttackDelay) && attacker.InterceptingSkillId == SkillId.None)
+				return CombatSkillResult.Okay;
+
+			var attackerPosition = attacker.GetPosition();
+			var mainTargetPosition = mainTarget.GetPosition();
+			if (!attacker.IgnoreAttackRange &&
+				(!attackerPosition.InRange(mainTargetPosition, attacker.AttackRangeFor(mainTarget))))
+			{ return CombatSkillResult.OutOfRange; }
+			if (!attacker.IgnoreAttackRange &&
+				(attacker.Region.Collisions.Any(attackerPosition, mainTargetPosition) // Check collisions between position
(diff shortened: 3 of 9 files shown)


exectails:
Thanks for taking the time to update your code. I was hoping you'd improve it a little though, it seems to still have the same problems I criticized last time, except for thankfully being more digestible now.

It feels very "hacked in". It's not very well commented, has long, complicated checks, deeply nested code, redundant code between and inside the handlers, new methods for existing behavior, and magic numbers. I've been starring at it for 20 minutes and still don't understand all changes, with some of them seeming pointless.

I won't be able to merge it as is, but I'll use your implementation as base. Thank you.

xKamuna:
I wasn't asked to clean it up so I didn't clean it up.  I only updated it to work with the latest aura.  I can fix it; it'd be easier if I knew what you wanted, though, through commenting on some lines.  Some things are dummy stuff that would have tied into Simultaneous Hits later if accepted, though, like CanAttack().  It isn't entirely redundant, as it would allow certain races/monsters not to windmill each other, and there hasn't been much found on how DevCAT allows/disallows that.

exectails:
Maybe you weren't explicitly asked, but I did tell you on the chat that I was hoping you would recode it, when you said you were planning to c&p the relevant code from your old PR^^"

>as it would allow certain races/monsters not to windmill each other

Well, we've debated that topic at length last time, and Aura currently does have a way of preventing monsters WMing each other. Speaking of WM, Windmill should probably not be called automatically by the handler via this system.

The wiki states that monsters Windmill automatically, and that players do it as well, but only if a certain *client-side* option is activated. In my opinion this means it's a client feature (normal and NPC client), where the client automatically determines that it should use Windmill.

Another redundant feature you added is the timer to make the creatures stand back up, which we also have, although it has a small flaw.

xKamuna:
Ah, is there something that makes the creatures stand back up?  When I was testing it, the creatures never stood back up unless they moved, so I added that in.
As for Windmill, I wasn't sure if it was client-sided for NPCs as well.
And as for everything, I have no idea what was added between my last commit and this one, so I just transferred everything unless I noticed something was changed or added.  I'll remove those three things.  What's the flaw in standing back up, by the way?  Perhaps I'll have some insight on it.

exectails:
The flaw is the same officials have, which is hard to fix. That they're potentially able to move before standing up, which makes them kinda stand up while already moving. A dedicated timer would lower the risk of that happening, but as I've said last time, I'm paranoid about having thousands of timers anyway, and adding more and more on each knock down doesn't improve the situation.

Instead I'm thinking about either making them stand up when a move comes in, or keeping them down until they're back up.

>When I was testing it, the creatures never stood back up unless they moved

Do you mean when you tested it last time, on your last PR? It has been over 1.5 years after all, hasn't it?

Sairii:
> Instead I'm thinking about either making them stand up when a move comes in, or keeping them down until they're back up. Though I just realized keeping them down is impossible because the client controls that =<

It is hard. Right now it feels really wierd and is very buggy on aura. Sometimes the enemies are downed forever and sometimes they teleport without any animation.
And its really rare to get that "shot em right after they downed and they get back up asap".
Im currently testing alot with ranged attack and keep on noticing that.

If it is possible to make them -nonglitchy then I would totally vote on leaving that "stand back up on move" out.
It should get up slowlyish on its own and not jump up like a madman and prolly teleport 10 meters.

exectails:
Don't confuse general glitches in the AI with the getting up issue though. I suspect there to be issues in the AI itself, which is why a rewrite is on the backlog.

xKamuna:
No, it was when I tested it on the current aura.
Perhaps it's not linked properly with stun time.
After all, there's new stun time business with the combat renewal.
I did notice that the timer method I used was rarely glitchy if ever on the old aura, though.

exectails:
Yes, because you're running a timer.

>Perhaps it's not linked properly with stun time.

It's not directly linked to the stun time at all, but to the IsStunned property. It will work on a current clean Aura for certain.

exectails:
@xKamuna It sounds like you would like to handle this, I can work on other stuff, but I'd like for it to get done soon. I mean, if it's working, there's not much to be done, remove 1-2 things and clean up a little, that's it.

Though I've been thinking about this system for the past few days, and there is one implementation detail that bothers me, the "attack out of range flag". It's imperative for your current way of calling the other creature's skill handler directly, but have you thought about either adding non-range-checked "overloads" or simply not allowing the attack to go through, and instead waiting for the other creature to come?

That way you have less random variables scattered around, that only serve to overcome a system limitation (this is what I mean by "hacky"), and maybe you can even find a way to get rid of the "intercepting skill" variable as well.

Race conditions are another point I'm worried about, and I've just realized this is not a problem on the single threaded official server... but I guess the risk is negligible.

xKamuna:
Well, I'm trying to keep the environment relatively D.R.Y.  How would these overloads be called?
On official, range is definitely ignored on simultaneous attack.  This is why you can attack a Giant Spider at the same time it attacks you.  Another example is a Golem using Smash, while you use Normal Attack.  So waiting for the other creature to come is not an option.  I don't think a few extra variables would hurt, unless multi-threading.  But "premature optimization is the root of all evil", so I think just noting it and coming back to it if we decide to use multi-threading even though the official servers don't would suffice.  But if you're very intent on such a thing, I can see what I can do to make it less variable dependent.  There are still some variables that need to be there that would not cause race conditions, such as the variable that signifies that a user has clicked and is ready for a simultaneous attack.  NotReadyToBeHit is also a timer that is necessary, not for simultaneous attacking, but in order to emulate the new behavior caused by the combat renovation where someone does not have all the properties of being "stunned" but still cannot be hit yet.

exectails:
>How would these overloads be called?

I don't know, `UseNoRangeCheck`?

>On official, range is definitely ignored on simultaneous attack.

I think you misunderstood what I meant. You have to ignore range to intercept hits the way you do it right now, like I said, but you don't need to ignore it for simultaneous attacks, since they can be handled directly, with no additional handler calling, and if you just don't let the first caller attack, and wait for the second attack to come in, in case the second is supposed to get the hit, you don't need to skip the check either. Or is there a reason why that's not possible?

>I don't think a few extra variables would hurt

If we said that every time, all classes would be riddled with one-time use variables though, and it would be a huge mess. I should mention that we actually have a class full of temp variables like that in Creature, CreatureTemp, which is used for situations where we couldn't get away without some kind of temp variable, that we didn't want to put into Creature. And that's only the ones we *really* needed.

>unless multi-threading

Aura is 100% multi-threaded, you are *never* safe, another call could come in at *any* time. Granted, in the case of packet handling the risk for two packets *from one creature* to be handled at the same time is basically non-existent, but from two creatures on the other hand is a real threat. I hope you didn't assume that wouldn't happen for your implementation.

>premature optimization

That's something very different. Premature optimization is bad because you're wasting valuable time squeezing performance out of something that doesn't necessarily need optimization, or because you're trying to save 1MB of RAM for a program that will run on 16GB RAM machines. We're not talking performance or resources here, we're talking code quality and safety, two very important concerns.

>There are still some variables that need to be there that would not cause race conditions

I'm not sure we mean the same thing in terms of "race conditions" here. What I'm referring to is the possibility that two creature's skill handlers run at the exact same time, which might very well happen, and the "second's" attack isn't stopped before damage is applied, packets are sent, etc.

>NotReadyToBeHit

Personally I would handle that in the on seconds tick the creature is subscribed to, but that variable will be removed sooner or later anyway. My running theory is that officials use the `Hit` lock for this. Official locks can be timed, are used for this kinda thing, and this lock doesn't seem to be related to hitting stuff yourself. The reason I haven't implemented it yet is that the lock system needs more research, I don't think our current implementation is 100% correct.

This will be needed for things like Windmill and Throw Paper Airplane invincibility as well.

xKamuna:
I meant how the overloads would be called, not what they would be called.  But now I'm confused on what you mean by overloads.  I thought you meant function overloads.

The simultaneous attack is calculated based on the first Creature that attacks.  Waiting is not an option because the second Creature's attack may not be called at all, either because they are out of range or they stopped their Normal Attack.  Remember that they don't have to be actually "attacking" to perform a simultaneous attack, they just have to have "clicked", have not gone out of combat mode, and are not using a skill that cannot simultaneously attack.  Which reminds me, I have to see if prepping a skill and canceling it also cancels the simultaneous attack state, as well as pressing the cancel skill button while doing a Normal Attack stops, I can't remember if I checked those.

I have to look again at my implementation to make sure I had multi-threading in mind.  If I didn't, I might have to get rid of those variables anyway.

>What I'm referring to
That's what I meant by race conditions as well.

>Official locks can be timed, are used for this kinda thing
Ah, okay, yeah, locks would be good for that, then.
 
Anyway, I'll clean up my code based on what you said and make sure there is no possibility of race conditions.

exectails:
>you need to explain the system sometimes

That's what I'm here for^^

>I meant how the overloads would be called, not what they would be called. But now I'm confused on what you mean by overloads. I thought you meant function overloads.

Ohh, well, I said "overloads", because I basically mean another method, that does what Use currently does, but without the range check. The range check would then be in the current Use method, which then calls the new method if the target check passes. That way it can be called either from the packet handler *with* range check, or without from wherever, if so desired.

>The simultaneous attack is calculated based on the first Creature that attacks. Waiting is not an option because the second Creature's attack may not be called at all, either because they are out of range or they stopped their Normal Attack.

You might be right... Out of range wouldn't be a problem, because you're waiting for them to come to you after all, but about them stopping their attack. I can see a way for this to work though, if the first person's attack isn't *canceled*, but just "*ignored*". Person A's request comes in first, server determines B is the one to get the attack, A's attack is not executed. Person A sends another request, server again determines B is the one to get the attack, A's attack is not executed, now B's request comes in, and they get the attack.

Problem with that is a potential endless loop though, unless we had a reliable way to determine who should get the hit. If we just used randoms, A's second request might actually get them the attack if we can't determine that it's still the same attack process.

I don't know how we would do that, but I feel like it would be a more robust solution, if we were able to reliably determine who will get the hit, without pure on-demand randomization. That's just a vague idea in my head though, I couldn't figure it out yet.

xKamuna:
The range check is done in Attack, though.  Passing it from Use to Attack would need a function variable in Attack, no?

Interesting idea, but I'm not sure that's the way to go about it.
In official, it really seems like the attack is cancelled and the opponent is forced to attack, even if they weren't going to send an attack request.  Again, if that weren't the case, we wouldn't be able to walk around and still intercept an attack, and out of range would still be a problem, because what if you're faster than them and you keep moving away?

exectails:
>so basically separate the current code, with the range check, and everything else after it in another?

Exactly. You could make it a parameter as well, but you would need a second method in any case, because Use is an interface method. Both work, I'd say it's a matter of preference.

>we wouldn't be able to walk around and still intercept an attack

You're right, forgot about intercepting Smash without actively running towards the target.

xKamuna:
Alright, I'll work on this pull request today and/or tomorrow, and you can check it out to see if the new format is good enough.

exectails:
The new structure is much better, even if the code is still a little confusing. While testing your PR two weeks ago I noticed a few things though.

1) Countering Smash doesn't work half the time. I haven't looked for the exact cause yet, but I assume the counter call doesn't work properly from one side.

2) Monsters are sometimes able to keep attacking after a knock back hit, where they should actually be stunned, and not be able to move. Even if you made sure they can't hit you until you recover, they will stay on your exact position, waiting for the millisecond in which they can hit again. This hasn't happened to me before, without your changes.

3) When thinking about how CM and Smash seem to be the only skills requiring this treatment, I started wondering if that's actually the case. I can't remember clearly, but I feel like I've seen a simultaneous Ranged Attack hit before. Am I mistaken here?

xKamuna:
1. Oh?  I'll have to look into that, then.
2. I believe that's an AI issue that has happened to me even without my changes.  I'll see what's up, though.
3. I'm pretty sure you're mistaken.

exectails:
Alright. Oh, and another thing, you checked for GM visibility in the skill handlers, but I don't think we should add that to every single skill handler, that would be very redundant, especially since there are central points where we could check it, the combat attack packet handler, the skill use packet handler, etc.

exectails:
@xKamuna Any updates?

xKamuna:
Yeah, I'm getting to it, school is getting heavy, so that along with my other projects, it is hogging my time.  I'll probably get to it next week, over Thanksgiving break.

exectails:
@xKamuna Any news? =<

xKamuna:
Don't worry, I'm working on it!

xCryptic:
Christmas break now? :^)

xKamuna:
I'm home now, I can now work on what I want at the pace that I wanted to.  So yes, Christmas break~

xKamuna:
Hm, I don't seem to have changed anything that would make countering Smash change at all.  I thought it would be the order switch doing such a thing, but that happens after counterattack.  Are you talking about intercepting Smash, not the Counterattack skill?

Issue #402, opened 2016-09-02
Collection books unimplemented


valdeza:
For instance, [Fashion Item - Hat](http://wiki.mabinogiworld.com/index.php?title=Collection_Books&oldid=619391#Fashion_Item_-_Hat) .

If research is necessary, I believe the easiest way to go out and gather packets to implement this would be to go to Voight and buy a collection book that can be completed by exploring around nearby.

### Useful Commands
cmd | desc
--- | ---
`>item 1501` | Spawn Fashion Item - Hat collection book
`>item 18022` | Spawn Leather Bandana, goes into book

exectails:
Right, those are implemented... did they exist in G1?

valdeza:
Can't really say. Mabiwiki didn't start keeping track of patches until G2 mid-2008

exectails:
We will need a log of putting an item into a book, and maybe also a log of how the reward is claimed. Though how the latter works might become obvious once we have the former. The progress is more than likely tracked via meta data, and it's quite possible that the reward marker can be identified by the name.

When an item is put into a book the server gets a request packet, it then has to check the book the item is added to, to see which items belong into that collection. This check, and what is supposed to happen besides adding to the book, like effects and notices, actually aren't all that simple, since there's a variety of collections.

There are several collection books that have special checks so we'll need scripting capabilities here. I suggest we add more methods to ItemScript, something like `OnCollect` and `OnReward`. Those would do nothing by default, but for the books we override them to do whatever needs to be done.

Examples of special behavior are collection books that require fishes of a certain size, items with specific suffixes or prefixes, items of a certain quality, etc. Unfortunately the collection db is server sided and can't be found in the client.

exectails:
Regarding the question of whether this is G1, I'm still not sure. I would think if they had had this feature in G2, they could've just used that for the Int Magic books, instead of abusing the enchanting system for it. But maybe that had different reasons.

exectails:
If someone could get a log of collecting an item in a book, and maybe even a completion packet, the actual implementation would be very easy.

DennouNeko:
Guess I could try to get some logs on that, once I get back home.
Voight should have some repeatable, relatively easy to complete collection books.

exectails:
Mabinogi CN's website has information about past updates, and while I don't read Chinese, the following image suggests collection books were added in G2.

[...]

PR #403, opened 2016-09-02
Add Simon PTJ, respective extra delivery quests

valdeza:
### Issues\_
- In this PR
  - Dialogue for the following PTJs have been improvised:
    - Extended delivery (Aeira -> Eavan)
    - Extended delivery (Walter -> Austeyn)
  - Details for the following extra quests have been improvised:
    - (id:40022) Deliver 10 Leather Bandanas
    - (id:40023) Deliver 5 Common Silk Weaving Gloves
    - (id:40024) Deliver 2 Cores' Healer Gloves
    - (id:40025) Deliver 1 Mongo's Hat
- To be solved in a separate PR
  - #402 Collection books unimplemented

ref: #237

### DIFF
--- a/system/scripts/quests/ptj/clothing_simon.cs
+++ b/system/scripts/quests/ptj/clothing_simon.cs
@@ -10,6 +10,7 @@ using Aura.Channel.Scripting.Scripts;
 using Aura.Channel.World.Entities;
 using Aura.Channel.World.Quests;
 using Aura.Shared.Util;
+using System;
 // Temporarily place this file under ChannelServer.csproj to resolve rest of dependencies
 
 // PR Readiness Check: 
@@ -33,6 +34,11 @@ using Aura.Shared.Util;
 // Simon's Clothing Shop Part-Time Job
 //--- Description -----------------------------------------------------------
 // All quests used by the PTJ, and a script to handle the PTJ via hooks.
+//--- Notes -----------------------------------------------------------------
+// Definitions for the following base scripts have been improvised,
+// update with official script whenever possible:
+// * SimonExtDeliveryAeiraEavanPtjBaseScript
+// * SimonExtDeliveryWalterAusteynPtjBaseScript
 //---------------------------------------------------------------------------
 
 public class SimonPtjScript : GeneralScript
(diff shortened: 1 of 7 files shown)


system/scripts/quests/ptj/clothing_simon.cs
```diff
@@ -0,0 +1,1487 @@
...
+	/// </remarks>
+	private int GetPersonalReportTime(Creature player)
+	{
+		Quest quest = player.Quests.GetPtjQuest();
+		if (quest == null)
+		{ // This should not normally happen.
+			Log.Error("SimonPtjScript: Player {0} does not have a PTJ report time for Simon. Used fallback of 1 AM.", player.Name);
+			return 1; // Fallback
+		}
+		else return quest.Data.ReportHour;
```
exectails: Maybe we should always use this instead of the const o_o

system/scripts/quests/ptj/clothing_simon.cs
```diff
@@ -0,0 +1,1487 @@
...
+	protected override int QuestId { get { return 510461; } }
+	protected override QuestLevel QuestLevel { get { return QuestLevel.Adv; } }
+}
+
+public abstract class SimonDeliveryStewartPtjBaseScript : SimonDeliveryPtjBaseScript
+{
+	protected override string LQuestDescription { get { return L("I'd like to give some [clothes] to Stewart at the School as a present. Can you give me a hand? - Simon -"); } }
+	protected override string LObjectiveDescription { get { return L("Deliver new [Clothes] to Stewart."); } }
+	protected override int ItemId { get { return 70038; } }
+	protected override string LItemNotice { get { return L("You have given Ruffled Tuxedo Ensemble to be Delivered to Stewart."); } }
```
exectails: I've lately been wondering if we should add a method that handles this notice automatically, together with removing the item. Like `player.GiveItem(npc, ...);`. The notice is in a standard format, "You have given X to Y.", so that's probably what officials do. But doing it manually might make translations more accurate o~o

valdeza: Would this be where Localization.GetParticular() could be used? We could get a translated format string through
```csharp
LXN(
	"{0} = Item Name(s), {1} = NPC Name",
	"You have given {0} to {1}.", "You have given {0} to {1}.", ItemCount
);
```
(We do have to consider cases where multiple items need delivering, such as the collection quests in this PR. The Deliver() objective also needs to be modified to take an ItemCount parametre.)

And how would we get localised names for the parametres? 
AuraData.ItemDb has the singular forme for ItemName, but not the plural. 
With NpcIdent, I'm not sure whether translating "_simon" to "Simon" occurs client-side or server-side, but if it's the former we would either need to somehow reference that or put all our NPCs through L() calls somewhere.

exectails:
>Would this be where Localization.GetParticular() could be used?

LXN is Localization.GetParticularPlural, but I'm not sure what you're trying to do there. Unless you try to abstract it, the messages will always be static, "give x of y to z". But yes, if you do want to abstract that, you would use GetPlural (LN).

Item names are tricky, because the current gettext parsers I found don't support JSON yet, so we can't just run it over the item db. Also, as you've mentioned, plurals are missing. My plan is to generate a C# file that doesn't really to anything but having references to the item name strings, so they can be parsed from there and we can get them localized, but we might never get a proper plural localization, because that would mean someone would have to regularly pluralize all items, old and new.

The NpcIdent isn't actually send to the client, or at least it shouldn't be. The client doesn't do anything with that, you have to handle that on the server. But L("Simon") shouldn't be a problem?

Don't go overboard with the abstraction, when I looked at the Aeira PTJ earlier for the supposed Eavan bug I really noticed how hard it is to follow what's going on^^" And ideally we would want newbies to be able to edit this stuff after all.

valdeza:
When Poedit loads an LX() string, whatever you specified as 'context' is shown to the translator above the usual Source/Translation text boxes. Aside from differentiating between words with multiple meanings, I also thought it could be used to inform the translator of what we expect {0} and {1} to be.

And sorry about the way I wrote Aeira's PTJ script. I was sort of trying out different abstractions in preparation for tackling... Stewart's PTJ...

image

Turns out there a lot of tiny variations with different quest IDs.
Also discovered some new ones while I was gathering Stewart's PTJ dialogue.

exectails:
>I also thought it could be used to inform the translator of what we expect

I see. It's not a bad idea, but I think abusing the context as a description might be misguided. Translators can easily look at the code where the localization call originated from within PoEdit, that should give them a general idea, as long as the variables have proper names.

Additionally, while we currently make no use of it, PoEdit, and the xgettext working in the background do support comments. By default comments prefixed with `TRANSLATORS:` should be picked up, as defined in the C# extractor settings for PoEdit. Though that might be a little heavy to have above all those L calls. Maybe only use it if it's really necessary. In most cases the meaning should be already inferable from the original text and the code.

Issue #230, opened 2016-02-07
Collection quests objective order


exectails:
Our quests are currently designed for sequential completion of objectives, so the next ones are revealed as you finish the previous ones.

- [X] Do something
- [ ] Do something else
- [ ] ???

Collection quests should work differently, in that you can finish the objectives in any order, with all of them being visible from the start. For example, the Black Orb collection quest, which you can finish in any order, with all fragment objectives being visible.

The quest *will* be completable once you have all items on Aura, but the objectives won't be visible or marked as done unless you have the first item, the second, the third, and so on. We could hardcode this behavior for Collect type quests, but I'd rather at a new property for it, defining how the quest should behave, in case we have to make a quest a Collect type quest (the type is sent to the client), even though it's not supposed to behave like this.

To accomplish this, the objective check logic has to be revised, to either check the current objective, or all of them, depending on the set behavior.

exectails:
Non-collection example: Party quest "Hunt Down the Dire Sprites", 3 kill objectives for all 3 sprite types, that don't have to be done in order.

exectails:
Since this will require some changes, we can look into properly fixing #390 on the way. Once an objective has been completed, we have to check if the next one might be done already, e.g collecting something, reaching a certain skill rank, etc.

valdeza:
Also, said "Collect type" quests should possibly remove items specified for each Collect quest objectives and not just the last ones ([a non-Collect objective could interrupt the item removal process](https://github.com/aura-project/aura/blob/08e42a5891cc8204802b26359344f64b88227f26/src/ChannelServer/World/Entities/Creatures/CreatureQuests.cs#L415)). 
Perhaps we should add an optional boolean parametre that flags an item to stay in inventory after completing a quest. Can't think of an example off the top of my head, but I get the feeling that would be needed in the future.

exectails:
That's not really part of this issue, but you're right of course. I did mention this somewhere, although I can't find it right now. Same for delivering. For this to work we need a callback for finishing objectives, which we currently don't have.

Ah, I remembered. I knew I mentioned it *somewhere*^^
https://github.com/aura-project/aura/blob/master/src/ChannelServer/Scripting/Scripts/QuestScript.cs#L1036

PR #404, opened 2016-09-05
Add Stewart PTJ

valdeza:
Note: Stewart does not have dialogue for speaking to him while on the job.

51 PTJs with small variations between them. I **had** to use that somewhat complex abstract class. D:

### Issue_ 
While still completable, progression with library PTJs is hindered with forced sequential objective completion and needing to trigger inventory update to acknowledge you collected a book. ( #230 )

_Resolved: Since the library is unavailable, shelving the books could not be tested. I **think** it works, though._

ref: #237

###### Sure the library is for the public, but the books are not for the public _to keep_. No one apparently returns books in Dunbarton!

_Edits_
1. Receive & shelve PTJs tested


valdeza:
Also (forgot to mention), Stewart doesn't exactly have quest levels. Someone qualified for advanced will get the same set of quests as one who is working for him for the first time.

To suppress that "Missing quest for level 'Adv'" warning, should we 
set all of Stewart's PTJs to QuestLevel.None and check for that when selecting a random PTJ or 
write a some method NpcScript.RandomPtjIgnoreLevel(PtjType, params int[]) that skips all the level checks?

exectails:
>To suppress that "Missing quest for level 'Adv'" warning

Actually, with this requirement change we could simply disable the warning. It's there because we were assuming same level PTJs being missing would be a bug, but if falling back to previous levels is an option if there are no higher ones, we don't need to warn anymore. It does what it's supposed to.

>I had to use that somewhat complex abstract class.

Technically you didn't *have* to* =P

>Since the library is unavailable, shelving the books could not be tested. I think it works, though.

It might have, but aside from the library warp not working, the bookshelf NPCs were also missing xD" Just added them, you can now test everything. Since it looks like they do nothing but End the dialog with a message right away by default, I gave them an `after_intro` hook, even though they don't really have an intro, and a call to End. If you hook them you should probably end the dialog from the hook, otherwise the script would continue to the End message.

xCryptic:
I didn't even know those bookshelves were g1 o_o

exectails:
Hm. Hard to say =/ The library dungeon is not G1, but did the library itself exist? If so, it's likely that the shelves existed as well, otherwise the region would've been empty.

exectails:
2000 lines, *with* heavy abstraction... and I was already annoyed by Endelyon's PTJ^^" Thanks for your hard work.

PR #415, opened 2016-09-09
Implement quest OnComplete(), #414 Fix - Stewart PTJ: Remove lingering books on quest completion

### LINKED ISSUE #414: Stewart's PTJ - multiple book quest
I collected books from Aeria, Walter and Simon to deliver them back to Stewart, but after completing the PTJ, Stewart only took one of the books (that being the one from Simon). The other two remained in my inventory. This guy must be the worst librarian in whole Erinn.

valdeza:
Quest objective order and types have been retained.

Last book to collect was skipped in this hotfix so as to avoid removing two books (if the player were to be holding onto two for some reason).

### DIFF
--- a/system/scripts/quests/ptj/library_stewart.cs
+++ b/system/scripts/quests/ptj/library_stewart.cs
@@ -357,18 +357,18 @@ public abstract class StewartVarLibraryPtjBaseScript : QuestScript
 
 	/// <summary>
 	/// Removes other quest items for PTJs involving more than one book.
-	/// <para>To be called manually.</para>
 	/// </summary>
-	/// <param name="player"></param>
+	/// <param name="creature"></param>
 	/// <remarks>
-	/// Hotfix #414: Stewart's PtjScripts manually removes items OnQuestComplete.
+	/// Fix #414: Lingering books on quest completion for Stewart PTJ
 	/// This might be removed when "any order" quest objective completion is implemented.
 	/// </remarks>
-	public void OnQuestComplete(Creature player)
+	public override void OnComplete(Creature creature)
 	{
-		// Skip last book; already taken care of by CreatureQuests.Complete(Quest, int, bool) .
+		// Remove all books except the last one;
+		// the last book is already taken care of automatically by CreatureQuests.Complete(Quest, int, bool) .
 		for (int i = NpcBookPairs.Length - 2; i >= 0; --i)
-			player.RemoveItem(NpcBookPairs[i].ItemId);
+			creature.RemoveItem(NpcBookPairs[i].ItemId);
 	}
 
 	private NpcScriptHook GenerateBookcaseAfterIntroHook(string objectiveIdent, int bookItemId)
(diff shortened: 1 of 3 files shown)


system/scripts/quests/ptj/library_stewart.cs
```diff
@@ -159,6 +159,13 @@ private void OnErinnMidnightTick(ErinnTime time)
 				return;
 			}
 
+			// Hotfix #414: Stewart's base PtjScript manually removes items OnQuestComplete.
```
exectails: What is this o,o Why not fix it properly?

valdeza: Oh, was I misusing the word 'hotfix'?

Currently:
Return to Stewart to report PTJ work. When reward is chosen, CreatureQuests.Complete(Quest, int, bool) will only remove the last book due to Talk objectives in between Collect objectives.

This PR:
Added method removes all quest book items except the last one. Aforementioned CreatureQuests.Complete(...) will only remove the last book. And so with this, all the books are removed once.

As for making OnQuestComplete(...) public, I was thinking it would be similar to how the OnReceive()  we have now was a virtual public method to be overriden as needed, if we do decide to have OnComplete() in the future. Named it OnQuestComplete() instead of OnComplete() because while this still compiles with child classes hiding inherited members, I was not absolutely sure that runtime compiler would be fine with it (just in case; from what I can tell, it seemed okay).

But if you prefer that I could add a Talk objective at the end to remove the books then, I could do that whenever I get home. I'll refactor OnQuestComplete(...) into the usual AfterIntro() we use for Talk objectives ~~and have it be called in StewartVarLibraryPtjBaseScript instead of StewartPtjScript~~. _edit: Oh no wait, it would have to be moved up to StewartPtjScript so that you give him the books_ before _automatically reporting to him about PTJs. Really, all that would be changed by doing this is an extra Talk objective at the end, but this offer still stands._

exectails: I see. It's not a bad idea, I don't like this fix because it's so hard to follow and unclean. Well, the whole script is... The completion callback should at least be implemented properly, then it would be okay for now =| An added bonus would be if the loop in OnQuestComplete isn't so hard to understand without having to read all comments and understand what's even going on.

valdeza: So to be sure I'm understanding you correctly, you would like me to do the following:

- Instead have OnComplete() as a virtual public method under the "Events" section in  Aura.Channel.Scripting.Scripts.QuestScript and call it as appropriate in  Aura.Channel.World.Entities.Creatures.CreatureQuests .
- Condense my above rationale into remarks for OnQuestComplete(...) .

exectails: There's not much need for a two paragraph long explanation on why it might be useful to be able to do something upon quest completion imo, but other than that, yes^^

valdeza:
Good golly, that's a long title.

Anyway, I hope this is more to your liking (or, at least a little bit). 
Tested it really quick by picking up books from Aeira, Walter, and Simon. Functions the same.

exectails:
A little, yes. More fix than hot-fix now^^ Thanks.

PR #418, opened 2016-09-12
Save all bank items

### LINKED ISSUE #417: Multiple Characters + Bank Tab "Dupe"
I hope I'm doing this right.

There's an issue with the way characters can access other character bank tabs.

If you store items into the bank on character A, then log onto Character B and withdraw the items from A's bank tab. Upon logging out and back in again onto character B, the items from character A will reappear for withdrawing again.

It sounds like a saving issue or something but, I could be wrong.

QewQew:
Fixes #417 by deleting all items on the character being saved along with all items in the bank belonging to the account before saving the items.
This adds a CreatureId to the bank tab in order to reference it when saving the tab.

### DIFF
--- a/src/ChannelServer/Database/ChannelDb.cs
+++ b/src/ChannelServer/Database/ChannelDb.cs
@@ -1,4 +1,4 @@
- // Copyright (c) Aura development team - Licensed under GNU GPL
+// Copyright (c) Aura development team - Licensed under GNU GPL
 // For more information, see license file in the main folder
 
 using Aura.Channel.Network;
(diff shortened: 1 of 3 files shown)


exectails:
@QewQew I just noticed a bug with this, saving *pets* throws and stops the entire saving process, since it's only deleting character items.

exectails:
Fixed in ea628b9e3ce4a8f9a82a25ee7486b40a209ec9b4. I basically reverted the deletion process and moved the saving to SaveAccount. This works and is even more performant.

PR #423, opened 2016-09-12
Fix removing option title "Speed" bonus

QewQew:
It was checking normal title, rather than the option title.


exectails:
Oh, thanks for the fix.

PR #428, opened 2016-09-14
Fix ErinnMonth starting at 0

QewQew:
Should have been changed along with f0d785772c15919970f2a0291b85ec6a7801a341

I was getting discounts today (Tuesday).


exectails:
Oh, thanks.

Issue #298, opened 2016-06-07
incorrect monster aggro


Pegwald:
_I didn't bother looking this up at the wiki since I have a feeling it might screw me over like it always does. I will follow my heart on this._

Some monsters have different AIs than what I was used to in the official version. I will list some mobs here:

-**Gorgon**: Should multi-aggro, but only aggro one by one.
-**Wood Jackals**: Just like Gorgon, they should multi-aggro if they are within aggro range.
-**Town rats**: Should **not** auto-aggro. I am not very bothered by this, but I don't remember them being so aggressive in the official version.
-**Werewolves**: Are they supposed to multi-aggro? 
-**semi-aggro>aggro**: I've noticed that if you are fighting enemies that don't multi-aggro, but aggro one by one, they will attack instantly if they go from a semi-aggro state to an aggro state, like when you kill an aggroed monster, the next mob will run at you immediately. Isn't there suppose to be some sort of delay before they aggro or something?
-**semi-aggro**: I might be wrong about this one, because I can't remember it quite right, but when you inflict damage in an AoE at several enemies, one will always aggro you, but aren't all the others supposed to become semi-aggro instantly?

exectails:
@Sairii, could you take a look at this? You've spend a lot more time with the AIs.

>Town rats: Should not auto-aggro

We only have one rat AI, which was probably written for the dungeon rats, which explains why they hate players. I guess we need a second one.

>Werewolves: Are they supposed to multi-aggro

No, I don't think so. Do they multi-aggro on Aura? o.o

>Isn't there suppose to be some sort of delay before they aggro or something?

Yea, there should be. Some tweaks in the AI have been postponed because the entire system is gonna get a rewrite soonish.

>aren't all the others supposed to become semi-aggro instantly?

Not necessarily, iirc that depends on the monster and the skill.

Pegwald:
> Do they multi-aggro on Aura?

I thought they didn't but suddenly I had 3 coming at me at once.
Here: https://www.youtube.com/watch?v=mx35--RvThY

exectails:
Yea, that can be attributed to the "not waiting" thing, they all 3 noticed that you're "free" at the same time and all started attacking, before one could "claim you".

Pegwald:
Ah, I see.

boilerplate5:
[...]

PR #436, opened 2016-09-17
AI's Update

Sairii:
Fixes for #298 
New AI's that were either missing or needed for the fix above.


system/scripts/ai/dungeonwolf.cs
```diff
@@ -0,0 +1,130 @@
+//--- Aura Script -----------------------------------------------------------
+// Dungeonwolf AI (Missing Wolf Support)
+//--- Description -----------------------------------------------------------
+// AI for WoodBoar,GiantJackal,WoodJackal,BlueDireWolf,Hardmode_BlueDireWolf,DarkRedDireWolf,Hardmode_DarkRedDireWolf,GiantWoodJackal
```
exectails: Since we don't have the race class names in our files, it would be better to use the actual names, otherwise someone might wonder what a DarkRedDireWolf might be. Comments should also be no longer than 80 characters, as per our coding conventions.

Sairii: oh whoops, well yeah thats true.
Yeah true it would be nicer to have our Names. But some of the Monsters are not in our DB yet.
DarkRedDireWolf = Burgundy q.q

exectails: Mhm, I looked them and their localized names up manually and fixed them. Though I was actually considering to put the class names in our db, since they're unique, while the localized names are often times the same. We have like 5 Light Gargoyles, who're all called Light Gargoyle, with little to differentiate them.

PR #447, opened 2016-09-26
Updated Tin

xCryptic:
• Updated Intro for #241 
• Added Greet
• Added Missing Keywords
• Added Gift Message
• Other small fixes

### DIFF
--- a/system/scripts/npcs/tin.cs
+++ b/system/scripts/npcs/tin.cs
@@ -30,9 +30,22 @@ public class TinScript : NpcScript
 
 		await Intro(L("A little boy with a heavy helmet is looking in my direction.<br/>The helmet is very well polished and features a dragon on the top, but prevents me from being able to see his face.<br/>He speaks in a low voice, and every once in a while places his left hand on his chin to keep his helmet on,<br/>as it slips off little by little."));
 
+		/* As of r234 (or earlier), you get this greet message on first time rebirth for 0 memory (memory over 0 is the same as normal rebirths)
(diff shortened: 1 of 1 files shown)


system/scripts/npcs/tin.cs
```diff
@@ -28,18 +28,11 @@ public override void Load()
...
-			"as it slips off little by little."
-		);
+		await Intro(L("A little boy with a heavy helmet is looking in my direction.<br/>The helmet is very well polished and features a dragon on the top, but prevents me from being able to see his face.<br/>He speaks in a low voice, and every once in a while places his left hand on his chin to keep his helmet on,<br/>as it slips off little by little."));
 
 		if (Player.Vars.Perm["EverRebirthed"] == null)
 		{
-			Msg("Hey, who are you?");
-			Msg("You don't look like you're from this world. Am I right?<br/>Did you make your way down here from Soul Stream?<br/>Ahhh, so Nao sent you here!");
-			Msg("She's way too obedient to the Goddess' wishes.<br/>Anyway, she's a good girl, so be nice to her.");
+			Greet();
```
exectails: Two calls to Greet are redundant, if you do it like this, you only need one call in front of the if, since it's always gonna be called. However, I'm unsure about the Greet method. You're doing the same check twice, instead of just handling the dialog here, which would make the code easier to follow.

exectails:
You forget the relation update, you would never get the second messages^^

xCryptic:
Oh... I did. That's weird, when I tested it, I swear I got the second message.

PR #450, opened 2016-10-02
Add Walter PTJ; Implement QuestRewardPattern

valdeza:
## Walter PTJ
Missing dialogue:
* First-time worker inquiry

Improvised dialogue:
* Extended item delivery: Glenis -> Aeira
* Extended item delivery: Nerys -> Stewart

ref: #237 

## QuestRewardPattern
Adv. PTJ to weave 5 Braid was discovered to reward a blacksmith manual. From what I can tell, we currently had no way to specify patterns for quest rewards. 
Also, I begin to wonder if I should've named my "Pattern" implementations as "Manual" instead.


valdeza:
Should we have another Item QuestReward that takes Item as a parametre instead of just item ID?
But if so, we have to be aware that
Method Aura.Channel.Scripting.Scripts.QuestScript.Item
conflicts in name with
Class Aura.Channel.World.Entities.Item

Scripters would have to do something like the following:
```csharp
using ItemEntity = Aura.Channel.World.Entities.Item;
// ...
public class NpcPtjBaseScript : QuestScript
{
	public override void Load()
	{
		// ...
		AddReward(1, RewardGroupType.Gold, QuestResult.Perfect, Item(ItemEntity.Create(/* ... */));
	}
}
```

valdeza:
And sorry for going dark. I had the PTJ done for a few weeks, but it needed testing and I didn't have a chance to add in and test QuestRewardPattern until tonight. Classes are really taking up a lot of my time.

As for what I'm planning to do next, I do recall promising a cleanup (localisation checks, refactoring Malcolm PTJ to use QuestScript.OnReceive(), etc.) of previous PTJs.

exectails:
A reward type that takes an actual item might not be a bad idea, but for now it's fine. Before we consider that, we would have to fix the ambiguous naming.

>I begin to wonder if I should've named my "Pattern" implementations as "Manual"

I wondered about that a few times myself. In some places we call it pattern, in some manual, it's tricky since it's basically the same exact thing internally, an item with a form id, but we're missing an overarching term =<

PR #458, opened 2016-10-15
Cleanup for current PTJs

valdeza:
Anything else that comes to mind?

===

### Change Summary
- Joined consecutive Msg() calls whenever possible.
- Removed one occurrence of a non-string, redundant Localization call.
- All logged warnings checked to state what codefile is printing the warning.  
  This helps when debugging rather than grepping for the warning message.
- Almost all logged errors de-escalated to warnings.  
  Errors perceived to be for more problematic situations. PTJ-caused warnings, although unexpected, have fallbacks.
- general_tracy.cs renamed to woodcutting_tracy.cs  
  This disassociates the file name from 'general store' classification. Quest script registration entry modified accordingly.
- Dialogue on delivery now correctly hides 'Name' rather than 'None'.  
  Auto-complete. Visually similar words. Easy to miss.
- In general_malcolm.cs, refactored method MalcolmPtjScript.GiveStartingPtjItems to instead override method QuestScript.OnReceive for each tailoring PTJ.

### Possibly Useful Regex Searches
- Non-string Localization calls  
  `/(?<=\W)LX?N?\((?!\s*")/g`
- Unlocalised string  
  (you will get a lot of false-positives without specifying exceptions)  
  `/(?<!(Log\.\w*|LX?N?)\(\s*)"(?!(AllowGroup))\w.*?"(?!\w)/g`  
  e.g. `(AllowGroup)` = `([@_]|ptj|after_intro|before_keywords|about_arbeit|ErinnMidnightTick|QuestViewRenewal|CollectionBooks)`
- Consecutive Localization calls(?)  
  Searches for `L("a") ...\n... L("b")` patterns that might be joinable like so: `L("a<p/>b")`  
  `/LX?N?\((.*?$\n.*?LX?N?\()+/gm`
- Parametrised strings - Just to check if Poedit can see your string  
  `/\.Format/g`


system/scripts/quests/ptj/general_malcolm.cs
```diff
@@ -1346,6 +1126,14 @@ public class MalcolmTailorPoposSkirtBasicPtjScript : MalcolmTailorPtjBaseScript
 	protected override string LCollectObjectiveDescription { get { return L("2 Popo's Skirts (F) (Part-Time Job)"); } }
 	protected override QuestLevel QuestLevel { get { return QuestLevel.Basic; } }
 
+	public override void OnReceive(Creature creature)
```
exectails: Hah, this is so much nicer~

exectails:
Thanks. About Tracy, in doesn't matter all that much, but the naming comes from the PTJ's category, and afaik Tracy is categorized as General (0).
https://github.com/aura-project/aura/blob/master/src/Mabi/Const/Quests.cs#L52

Every time I want to look up a specific PTJ I'm thinking we might want to change the naming though, since I have not once searched for a category so far, but always for the NPC name. Categories seemed useful, to find all church PTJs for example, but searching for that will rarely be the case I guess =|

PR #460, opened 2016-10-18
Add Elen PTJ, Big Order of Iron Ingots collection quest 

valdeza:
The following dialogue is missing: 
* First-time worker inquiry
* 1-star PTJ turn in response
* 3-star PTJ turn in response

ref: #237


exectails:
Must be nice to be done with the complicated PTJs^^

PR #461, opened 2016-10-19
(Testing Needed) Add Comgan PTJ

valdeza:
The following dialogue is missing: 
* First-time worker inquiry

ref: #237

### DIFF
--- a/system/scripts/npcs/bangor/comgan.cs
+++ b/system/scripts/npcs/bangor/comgan.cs
@@ -73,7 +73,11 @@ public class ComganScript : NpcScript
 
 	private void Greet()
 	{
-		if (Memory <= 0)
+		if (DoingPtjForNpc())
+		{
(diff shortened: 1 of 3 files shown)


valdeza:
At the time of writing, I seem to have network issues despite running both client and server on localhost.
Every server except Web gives me:

> `[Exception] - System.Net.Sockets.SocketException (0x80004005): Only one usage of each socket address (protocol/network address/port) is normally permitted`

I guess that's something to troubleshoot tomorrow-ish.

QewQew:
Make sure that nothing else is using a port required by aura and that no other servers are already open. Try restarting your computer if you're sure nothing is using the ports.

valdeza:
Normally, I only PR tested code, so I just thought it might be appropriate to give a reason?
That and we're kinda extremely close to closing that PTJ issue.

(Discussion on socket troubles continued in Gitter)

PR #499, opened 2016-12-05
Set SetCancelable to true for all fomor scroll quests

eotetb:
Like party quests, fomor scrolls should also have the ability to be canceled through their quest scrolls.


exectails:
Really? Huh. Thanks for the PR.

PR #510, opened 2016-12-11
Added Gift & Relation Messages to Duncan

xCryptic:


system/scripts/npcs/tir/duncan.cs
```diff
@@ -153,9 +158,28 @@ private void Greet()
...
+				{
+					Msg(FavorExpression(), "I should take care of the graveyard now.<br/>I need some time to prepare, so I won't have much time to converse for the rest of the day.");
+					ModifyRelation(Random(2), -Random(2), Random(1, 4));
+				}
 				else
 				{
-					if (Title == 33)
+					if (Title == 33 && Player.Age > 20)
+					{
+						Msg(FavorExpression(), "(Missing dialog: Information about the Diligent title while being over the age of 20).");
```
xCryptic: @valdeza Seeing you are doing ptjs, I would assume you got this title (title 33 is the Diligent, which you get after doing 100 ptjs). If your character matches the criteria for this `if`, could you get the message here? Thanks.

PR #519, opened 2016-12-15
Updated SuccubusAi

Sairii:
Updated the MagicStack Attack and Prepares.

### DIFF
--- a/system/scripts/ai/succubus.cs
+++ b/system/scripts/ai/succubus.cs
@@ -389,8 +389,6 @@ public class SuccubusAi : AiScript
 			if (Random() < 30)
 			{
 				Do(StackAttack(SkillId.Firebolt));
-				Do(Wait(1000, 2000));
-				Do(Attack(1, 4000));
 				Do(Wait(2000, 3000));
 				Do(Say("You're not mad at me, are you? (Laughter)"));
 			}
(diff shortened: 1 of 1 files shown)


system/scripts/ai/succubus.cs
```diff
@@ -683,21 +672,10 @@ protected override IEnumerable Aggro()
...
-						Do(Wait(500));
-					}
-					else
-					{
-						Do(PrepareSkill(SkillId.Lightningbolt)); // 2 charge
-						Do(Wait(1000, 2000));
-						Do(Attack(1, 4000));
-						Do(Wait(500));
-					}
+					Do(PrepareSkill(SkillId.Lightningbolt, Rnd(1, 1, 2)));
```
Sairii: Pretty much the same thing. I want that wait and with stackattack I dont have that.

system/scripts/ai/succubus.cs
```diff
@@ -399,7 +388,7 @@ protected override IEnumerable Alert()
 
 			if (Random() < 30)
 			{
-				Do(PrepareSkill(SkillId.Firebolt)); // Actually does only attack with previous charges, charges only if no charges
```
exectails: StackAttack does have a builtin 1-2 seconds wait time.

Sairii: actually no im wrong, i can do the wait before the stack attack

exectails:
Thanks for the update.

PR #520, opened 2016-12-16
Update Shiela RP to end after killing 3 bosses

eotetb:
Like in the Mores RP, the dungeon should end after killing 3 Ghost Armors rather than needing to kill all 6 bosses.


exectails:
Thanks!

PR #539, opened 2016-12-24
Edern PTJ: Adjusted player skill check

valdeza:
(5b86004 follow-up)

Also, Happy Holidays to all~

### DIFF
--- a/src/ChannelServer/Scripting/Scripts/NpcScript.cs
+++ b/src/ChannelServer/Scripting/Scripts/NpcScript.cs
@@ -1346,13 +1346,13 @@ namespace Aura.Channel.Scripting.Scripts
 		}
 
 		/// <summary>
-		/// Returns a random quest id from the given ones, based on the current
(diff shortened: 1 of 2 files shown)


exectails:
Could you elaborate on what exactly you changed?

valdeza:
> We either have to adjust the check or add the potentially missing PTJ.

And I did the former:
- If we are not able to find appropriate PTJs for certain Blacksmithing skill rank/PTJ level combinations (e.g. [[this case]](https://github.com/aura-project/aura/commit/5b86004d23d79d7e4c5ad139bba21395e239b125#commitcomment-20208628)), the check now retries on lower skill ranks.
- If we still failed to find something, the script now reports the problematic combination for debugging. Though it seems to work for all the possible combinations I know of.
- I removed the extra quest we inserted as a remedy; we're now "official" until we can confirm all the quests Edern offers.

exectails:
Thanks.

PR #544, opened 2016-12-27
Added Aer

eotetb:
Adds npc Aer to the underground spring at Ceo Island.

### DIFF
--- a/system/scripts/npcs/ceo/aer.cs
+++ b/system/scripts/npcs/ceo/aer.cs
@@ -1,0 +1,174 @@
+//--- Aura Script -----------------------------------------------------------
+// Aer
+//--- Description -----------------------------------------------------------
+// Water spirit in the underground spring at Ceo Island
+// Gives a Rundal Siren Pass when gifted with a Suspicious Fomor Pass
+//---------------------------------------------------------------------------
+
+public class AerScript : NpcScript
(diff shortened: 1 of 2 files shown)


system/scripts/npcs/ceo/aer.cs
```diff
@@ -0,0 +1,174 @@
...
+					"Sorry... I don't know anything about that."
+				);
+				ModifyRelation(0, 0, Random(3));
+				break;
+		}
+	}
+
+	protected override async Task Gift(Item item, GiftReaction reaction)
+	{
+		if (item.Info.Id == 63103) // Suspicious Fomor Pass
```
exectails: You should add a feature check for `RundalDungeon` here, so nobody can get the pass this way through potential bugs, glitches, hacks, etc. unless it's enabled.

eotetb: I went ahead and added a feature check to that part of the script. However, Rundal Siren should probably have its own separate feature check. It wasn't added until g2s3 according to both the wiki and Nexon's own patch notes. 

source: http://mabinogi.nexon.net/News/All/1/000aY

exectails: Good catch, I missed that. Thanks.

PR #564, opened 2017-01-03
Adds Emain Macha Royal Guardsmen

eotetb:
Adds the royal guardsmen to Emain Macha.


system/scripts/npcs/emain/guardsman.cs
```diff
@@ -0,0 +1,155 @@
...
+	
+	protected override async Task Talk()
+	{
+		RndMsg(
+			"My back hurts",
+			"My feet hurt...",
+			"No talking is allowed during duty",
+			"I cannot talk to you while on duty!",
+			"...We're the proud guards of Emain Macha!<br/>As long as we're here, there's no need to fear!",
+			"(A guard with shining armor is standing post.)Everything here is in order!!", //hide name
```
exectails: You can hide the name by either not using RndMsg, but creating a switch with normal Msg calls,
```csharp
switch (Random(2))
{
    case 0: Msg("..."); break;
    case 1: Msg(Hide.Name, "..."); break;
}
```
or you can use XML.
```xml
"<title name='NONE' />(A guard with shining [...]
```

system/scripts/npcs/emain/guardsman.cs
```diff
@@ -0,0 +1,155 @@
+//--- Aura Script -----------------------------------------------------------
+// Emain Macha Royal Guards
+//--- Description -----------------------------------------------------------
+// Script for the royal guard npcs in Emain Macha
+// guardsman07 is different from the other guards and says lots of gossip
+//---------------------------------------------------------------------------
+
+public class _GuardsmanBaseScript : NpcScript
```
exectails: If you don't want a script to be initiated you should mark it as abstract, and there's no reason for the underscore.
```csharp
public abstract class GuardsmanBaseScript : NpcScript
```

system/scripts/npcs/emain/guardsman.cs
eotetb: I wondering how to go about doing that, thanks.

exectails:
Thank you for contribution!

PR #569, opened 2017-01-04
Added Dual Gun Skills

xeroplz:
- Adds all Dual Gun skills
- Adds bulletsMax entries in the item database for guns
- Adds partial support for MultiHit TargetOptions

Post Cleanup 01.03.2017
- Skill Helpers added
- Gun tags and bullet count added to Creature
- Unfinished Reforms: DGM Crit Chance, Target Area Polygon SkillHelper,
ShootEffect Packet optimization, ToList methods in
BulletStorm/ShootingRush


src/ChannelServer/Skills/Guns/BulletSlide.cs
```diff
@@ -0,0 +1,316 @@
...
+
+			// Damage
+			var damage = (attacker.GetRndDualGunDamage() * (skill.RankData.Var2 / 100f)) * tAction.MultiHitCount;
+
+			// Master Title
+			if (attacker.Titles.SelectedTitle == skill.Data.MasterTitle)
+				damage += (damage * (15 / 100f)); // +15% damage
+
+			// Critical Hit
+			var dgm = attacker.Skills.Get(SkillId.DualGunMastery);
```
xeroplz: @exectails I don't know how this could be incorporated into a helper, since it needs to actually return a value. It definitely wouldn't fit as a static value in SkillHelper

xeroplz: @exectails I don't know how to fit this into a helper. It definitely can't be a static value in SkillHelper, because otherwise it wouldn't be disposed of... right?

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,461 @@
...
+			{
+				targets = targets.Take(maxTargetsBulletCount).ToList();
+			}
+			else if (maxTargetsBulletCount >= skill.RankData.Var5)
+			{
+				targets = targets.Take((int)skill.RankData.Var5).ToList();
+			}
+
+			// Effects & Etc
+			var shootEffect = new Packet(Op.Effect, attacker.EntityId);
```
xeroplz: What you said about creating a list with the distances and target ids, I really don't understand how that would go about :L

src/ChannelServer/Skills/Guns/BulletStorm.cs
```diff
@@ -0,0 +1,461 @@
...
+			var cap = new CombatActionPack(attacker, skill.Info.Id);
+
+			var aAction = new AttackerAction(CombatActionType.SpecialHit, attacker, targetAreaId);
+			aAction.Set(AttackerOptions.UseEffect | AttackerOptions.KnockBackHit1);
+			cap.Add(aAction);
+
+			aAction.Stun = AttackerStun;
+
+			// Get targets in descending order for effect packet using bulletDist
+			var targets = attacker.Region.GetCreaturesInPolygon(p1, p2, p3, p4).Where(x => attacker.CanTarget(x) && !attacker.Region.Collisions.Any(attackerPos, x.GetPosition())).OrderBy(x => x.GetPosition().GetDistance(attackerPos)).ToList();
```
xeroplz: removing ToList() from targets caused issues with lines 172-179, and honestly, i have no clue what to do there 😄 
ss 2017-01-03 at 06 37 33

PR #570, opened 2017-01-05
Added Nekojima Field Spawns

eotetb:
Adds field spawns to the southern island in Nekojima.

### DIFF
--- a/system/scripts/regions/nekojima.cs
+++ b/system/scripts/regions/nekojima.cs
@@ -1,0 +1,43 @@
+//--- Aura Script -----------------------------------------------------------
+// Nekojima
+//--- Description -----------------------------------------------------------
+// Warp and spawn definitions for Nekojima
+// Also includes a global drop for monsters with the /rat/ tag 
+// for the cat bell item while in the Nekojima region
+//---------------------------------------------------------------------------
+
+public class NekoRegionScript : RegionScript
+{
+	public override void LoadSpawns()
+	{
+	//Nezumijima spawn group 1
+	CreateSpawner(race: 120002, amount: 6, region: 600, coordinates: A(88898,72506, 90125,69636, 92898,71522, 91173,73342)); // Brown Town Rat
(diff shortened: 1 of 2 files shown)


system/scripts/regions/nekojima.cs
```diff
@@ -0,0 +1,43 @@
...
+	//Nezumijima spawn group 2
+	CreateSpawner(race: 170301, amount: 4, region: 600, coordinates: A(92854,72980, 93426,70564, 93788,71873, 93802,72804)); // Ratman
+	CreateSpawner(race: 120010, amount: 1, region: 600, coordinates: A(92854,72980, 93426,70564, 93788,71873, 93802,72804)); // Black Ship Rat
+	}
+}
+
+public class CatsBellDropScript : GeneralScript
+{
+	[On("CreatureFinished")]
+	public void OnCreatureFinished(Creature creature, Creature killer)
```
eotetb: I attempted to originally, but I had trouble checking for region id with that feature.

exectails: You would have needed to create a new GlobalDrop class from your script, something we talked about on the chat recently, but after reading this I also added a region GlobalDrop class to the core earlier.

exectails: It seemed weird to me that they would make a global drop out of this, instead of creating copies of the monsters or something like they usually do, so I just took a look after merging. I can't find a reference to those rats dropping that item right now. The rat field bosses are apparently supposed to drop those Cat items, but are the normal spawns as well?

eotetb: Yeah, they are supposed to. I actually tested this with other rat monsters that don't normally spawn in the region and they also dropped bells when in the region. I tested non rat monsters and they did not drop bells. This was with the g13 server files.

exectails: I see, gj.

PR #573, opened 2017-01-06
Added Produce Massive Elemental Quests

eotetb:
Adds the three produce massive elemental quests.


exectails:
Thank you! The wiki says those are sold by Stewart, are you going to add them to the shop as well?

eotetb:
On another note, the massive elementals themselves need to be implemented #575.

PR #577, opened 2017-01-07
Added Nele

eotetb:
Adds Nele to Emain Macha.

### DIFF
--- a/system/scripts/npcs/emain/nele.cs
+++ b/system/scripts/npcs/emain/nele.cs
@@ -1,0 +1,487 @@
+//--- Aura Script -----------------------------------------------------------
+// Nele
+//--- Description -----------------------------------------------------------
+// Bard and music item shop-keeper
+//---------------------------------------------------------------------------
+
+public class NeleScript : NpcScript
+{
+	public override void Load()
+	{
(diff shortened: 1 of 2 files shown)


eotetb:
The secret shop and the pre-made score scrolls still need to be implemented. 

Unlike most other secret shops, Nele's adds books to a shop tab that already exists rather than adding a new one.

system/scripts/npcs/emain/nele.cs
```diff
@@ -0,0 +1,487 @@
...
+					Msg("I can't believe you, the one who was researching about Macha earlier,<br/>has become the Guardian of Erinn...");
+					Msg("...I'm assuming you've stopped Macha's curse...?<br/>Thank you. <username/>...");
+				}
+
+				await Conversation();
+				break;
+
+			case "@shop":
+				if (Memory >= 15 && Favor >= 50 && Stress <= 5)
+				Msg("Hey, are you here to buy something from me again?<br/>Since you are here so often, <username/>, I feel like I have to show you something very special...<br/>Haha, take a look...");
```
exectails: You should format your code properly, these Msgs go one indent deeper.

PR #580, opened 2017-01-08
Added Bastet

eotetb:
Adds the General Shop npc, Bastet to Nekojima.

### DIFF
--- a/system/scripts/npcs/nekojima/bastet.cs
+++ b/system/scripts/npcs/nekojima/bastet.cs
@@ -256,18 +256,18 @@ public class BastetScript : NpcScript
                 else if (Favor <= -10)
                 {
                     RndFavorMsg(
-                    "...Umm...",
-                    "Huh? Did I hear that wrong?",
-                    "I don't understand to well..."
+                   	"...Umm...",
+                    	"Huh? Did I hear that wrong?",
+                    	"I don't understand to well..."
                     );
                     ModifyRelation(0, 0, Random(4));
                 }
                 else if (Favor <= -30)
                 {
                     RndFavorMsg(
-                    "...Umm...",
-                    "Huh? Did I hear that wrong?",
-                    "I don't understand to well..."
+                 	 "...Umm...",
+                  	 "Huh? Did I hear that wrong?",
+                  	 "I don't understand to well..."
                     );
                     ModifyRelation(0, 0, Random(5));
                 }
--- a/system/scripts/npcs/nekojima/bastet.cs
+++ b/system/scripts/npcs/nekojima/bastet.cs
@@ -15,12 +15,10 @@ public class BastetScript : NpcScript
         SetLocation(600, 102115, 97502, 20);
         SetGiftWeights(beauty: 0, individuality: 2, luxury: 1, toughness: 1, utility: 0, rarity: 1, meaning: 2, adult: 1, maniac: 1, anime: 1, sexy: 0);
 
-
         EquipItem(Pocket.Face, 3908, 0x00F79723, 0x0057686A, 0x0005004F);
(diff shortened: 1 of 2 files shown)


eotetb:
Hopefully the English scripts I 'translated' are decent enough, otherwise I do still have a copy of the script with the original Japanese text if anything needs to be redone.

The npc's name is also set to `_<mini>NPC</mini> Bastet` because using `_bastet` gives us her name in Korean with the English localization files.

system/scripts/npcs/nekojima/bastet.cs
```diff
@@ -0,0 +1,348 @@
...
+                }
+                else
+                {
+                    Msg(FavorExpression(), "Have you seen the rats on the island beyond the water bridge?<br/>I went for a stroll without knowing,<br/>anything and then a rat came out...<br/>I thought my heart was going to stop! After that I will not go there by myself.<br/>Sheesh... I'm horrified just thinking about it.");
+
+                    ModifyRelation(Random(2), 0, Random(3));
+                }
+                break;
+
+            case "about_arbeit":
```
exectails: Does the NPC even have a PTJ? If not there's not reason to say that it's unimplemented.

exectails:
You might want to use `SetDefaultName` to get rid of the mini NPC in the title of the dialog box if you use that code to get the NPC tag. Something to also note when you do this is that you lose the portrait if the NPC has one. Though I guess the Nekojima NPCs might not.

system/scripts/npcs/nekojima/bastet.cs
eotetb: This npc actually does have a PTJ, as do the other Nekojima npcs. I will add those some time later.

eotetb:
Yes, none of the npcs have portraits. `SetDefaultName` would have to be applied to every Msg line correct?

exectails:
No, that's why it's called "Default"Name =)

eotetb:
It didn't seem to do anything at all when I added `SetDefaultName("Bastet");` at the very beginning of talk (it still showed up with the npc tag in the dialogue box). 

Also, I did find a way to call portraits using XML, for example,
`<npcportrait name='lorna'/>`
Unfortunately it seems you must use that every single line, but it appears that's what Nexon did for Price/Redire to change his portrait after beating G2. There was also a function to change npc facial expressions as well.

exectails:
You can actually use SetPortrait, which then appends the portrait code automatically.

eotetb:
When I attempted to use SetPortrait in the past it didn't seem to do anything (with lorna specifically). Although, using her official npcName didn't cause it to show up either, so perhaps that was related to SetPortrait not working. I suppose this is a little off topic from this issue though.

exectails:
Ah, my mistake, it's `Msg(SetDefaultName("..."));`. *That* works. Though it's not really straight forward I guess, we might want to work on that. You can use it though. And SetPortrait should work fine, it does exactly what you would do manually.

```csharp
SetPortrait("lorna");
Msg(SetDefaultName("Foobar"));
Msg("Sup?");
```

ss 2017-01-08 at 11 37 19

eotetb:
That works, I pushed the change.

I wonder why I had issues with the npc portraits then if it seems to be working just fine for you though...

system/scripts/npcs/nekojima/bastet.cs
```diff
@@ -0,0 +1,347 @@
...
+        EquipItem(Pocket.Shoe, 17021, 0xFFFFFFCC, 0x00000000, 0x00000000);
+
+        AddPhrase("I'd like some new clothes to wear~!");
+        AddPhrase("I wonder if I look young~?");
+        AddPhrase("I don't feel like I need a diet yet... Probably...");
+    }
+
+    protected override async Task Talk()
+    {
+    	Msg(SetDefaultName("Bastet"));
```
exectails: You didn't use tabs.

PR #589, opened 2017-01-11
Golden Ratman Field Boss

eotetb:
Adds the giant golden ratman field boss to Nekojima. Also adds both the giant golden ratman race and dark ratman race along with magic ratman ai (for the dark ratman minions).


exectails:
How do you create these pull requests? My client always tells me that the whole file was changed as soon as I merged it, which indicates that the line breaks are incorrect.

exectails:
Oh, and I just realized that didn't include the boss drops for the Giant Golden Rat Man. It's supposed to drop those cat items and stuff, isn't it?

https://github.com/aura-project/aura/blob/master/system/scripts/events/pillow_fight.cs#L264

eotetb:
I didn't realize we had a feature like that actually, the drops were temporarily included on the race (although apparently that monster doesn't actually have drops, just the boss drops). 

I normally use notepad ++ instead of a full IDE like Visual Studio to create these pull requests, mainly because I'm more familiar with it and I find it easier to reference files using the split view the program has, but after I'm done in there I've been using Visual Studio's auto format with the last few pull requests, but it seems like that has been causing more problems rather then fixing them.

exectails:
The IDE should be irrelevant, the git client is supposed to normalize the line breaks on commit. That is unless it's configured to not do that, or you're not using git.

eotetb:
Actually I've never tried using the Git Client, perhaps I should try using that.

PR #597, opened 2017-01-13
Add Artisan Upgrades

valdeza:
_Implements #134_

Oh, these. The couple of times I tried back in the day, I remember getting terribad rolls on both. :|


src/ChannelServer/Scripting/Scripts/NpcScript.cs
```diff
@@ -1175,10 +1175,9 @@ public UpgradeResult Upgrade(string upgradeReply)
 			}
 
 			// Check for disabled Artisan
-			// TODO: Feature check, once we do have Artisan.
-			if (result.Upgrade.Effects.Any(a => a.Key == "Artisan"))
+			if (result.Upgrade.Effects.Any(a => a.Key == "Artisan") && !IsEnabled("ArtisanUpgrades"))
 			{
-				Send.MsgBox(this.Player, Localization.Get("Artisan upgrades aren't available yet."));
+				Send.MsgBox(this.Player, Localization.Get("Artisan upgrades are currently unavailable."));
```
exectails: I made it "not yet", because "currently unavailable" is ambiguous. It could mean that they're being worked on, or that they were temporarily disabled, but this case means that they aren't available yet, that the feature is disabled, probably because the server isn't in that generation yet.

system/db/features.txt
```diff
@@ -541,6 +541,9 @@
 		
 		// Makes sheep protection quest easier. (Part of gfC4Tutorial)
 		{ name: "ShortSheepProtection", enabled: true },
+		
+		// Enables artisan upgrades
+		{ name: "ArtisanUpgrades", enabled: true },
```
exectails: The official feature is called "ArtisanUpgrade", without "s" at the end.

exectails:
Nice job!

src/ChannelServer/Scripting/Scripts/NpcScript.cs
valdeza: I.. never really considered the implications.
To me it just looked more formal to have the server say that instead.

Anyway, they both imply that Artisan upgrades will be available _some time_ in the future. But the original statement seems to be more explicit in promising that it will be enabled, while the revision intentionally leaves that unclear. For instance, consider a server that is at least on that generation and Artisan upgrades were enabled for a time, to be disabled later: the revision would appear to fit that scenario better.
Ambiguous wording would have this statement fit a wider spectrum of scenarios on how/why Artisan upgrades are not enabled. Though much like a fortune cookie clearly tells you "something is going to happen in the future, who knows what exactly", the primary message here is "Artisan upgrades are disabled, not sure until when".

src/ChannelServer/Scripting/Scripts/NpcScript.cs
```diff
@@ -1355,9 +1354,46 @@ public UpgradeResult Upgrade(string upgradeReply)
...
+						var artisanData = AuraData.ArtisanUpgradesDb.Find(artisanId);
+						if (artisanData == null)
+						{
+							Log.Warning("NpcScript.Upgrade: Upgrade '{0}' references unknown Artisan upgrade '{1}'.", result.Upgrade.Ident, artisanId);
+							break;
+						}
+
+						// Choose a random number of random effect sets to apply to item.
+						// Has to be at least one of the random effect sets.
+						var randomEffectSet = UniqueRnd(Random(artisanData.Random.Count) + 1, artisanData.Random.ToArray());
```
valdeza: And while we decide whether or not we revert the rewording of "Artisan upgrades are disabled,"
let's not forget to check whether `artisanData.Random.Count` is 0, because if so, an exception would occur.

src/ChannelServer/Scripting/Scripts/NpcScript.cs
exectails: Maybe we should get rid of the "currently" and "yet", removing any promise? I mean, what if someone doesn't want Artisan upgrades, and keeps them disabled, in that case they'll never be there and are simple "unavailable".

system/db/artisanupgrades.txt
```diff
@@ -0,0 +1,153 @@
+// Aura
+// Database file
+// 
+// As of 10 Jan 2017, the values for the following IDs are unknown:
+// 124, 125, 131, 133, 134, 136, 138, 139
+// 
+// Values: id, [random[]], [always[]]
+// 
+// Option set IDs defined in 'random[]' will randomly have at least one effect applied on the equipment.
+// Option set IDs defined in 'always[]' will always be applied on the equipment.
```
exectails: Just noticed that there's only one `always` array in the entire db, is that correct?

valdeza: As far as I know. That entry is for [Artisan upgrades on magic armour](http://wiki.mabi.world/index.php?title=Armor_Upgrades/Magic&oldid=657751#cite_ref-3). I don't know of any other instances where you have guaranteed stats.

exectails: Alright, thanks!

PR #598, opened 2017-01-13
Added 'prof' Command

valdeza:
This command sets equipment proficiency.

Came in handy when testing the Artisan upgrades.

### DIFF
--- a/src/ChannelServer/Util/GmCommands.cs
+++ b/src/ChannelServer/Util/GmCommands.cs
@@ -96,7 +96,7 @@ namespace Aura.Channel.Util
 			Add(50, 50, "keyword", "[-|+]<name>", Localization.Get("Adds/removes keywords."), HandleKeyword);
 			Add(50, 50, "ptj", "<type> <level>", Localization.Get("Sets the level of a certain PTJ type."), HandlePtj);
 			Add(50, 50, "quest", "<id> [finish objective]", Localization.Get("Starts a quest or sets the specified objective to be finished."), HandleQuest);
-			Add(50, 50, "prof", "[equipflags=7 (1:armor,2:rhand,4:lhand)] [amount=101000]", Localization.Get("Sets proficiency of equipped item(s)."), HandleProf);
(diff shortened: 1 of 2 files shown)


src/ChannelServer/Util/GmCommands.cs
```diff
@@ -2405,6 +2406,49 @@ private CommandResult HandleQuest(ChannelClient client, Creature sender, Creatur
...
+
+			if (args.Count >= 3 && !int.TryParse(args[2], out profAmount))
+			{
+				Send.ServerMessage(sender, Localization.Get("Could not parse 'amount' argument."));
+				return CommandResult.InvalidArgument;
+			}
+
+			var targetEquipment = target.Inventory.GetMainEquipment(item =>
+				((equipFlags & EquipFlag_Armor) != 0 && item.Info.Pocket == Pocket.Armor) ||
+				((equipFlags & EquipFlag_RHand) != 0 && item.Info.Pocket == (target.Inventory.WeaponSet == WeaponSet.First ? Pocket.RightHand1 : Pocket.RightHand2)) ||
```
exectails: There's a shortcut for that pocket check, `target.Inventory.RightHandPocket` will return the correct pocket based on the weapon set. Same for the left hand of course.

exectails:
I like the idea, but I find the bitmask a little unwieldy. I wonder if there might be a better way to go go about that, especially since someone might want to prof other items as well.

valdeza:
You mean like proffing an item that they cannot equip? I'm not sure if the (DEV) text that appears when hovering over an item is client-side or server-side, but a clear catch-all solution here would be to pass in item entity IDs instead. 
But that would be expecting the command user to be packet sniffing for that entity ID that, as far as I know, cannot normally be shown in the client. 
Either that or be running a modded client that will display that entity item ID in the hoverbox. Trivial or not, I'm not sure; I have never made my own mods.

But if we are to stick to only equipped items, we could instead have the user name the slot to prof (e.g. "armor", "glove", "shoe", "lhand", "altrhand"; [similar to the 'go' command](https://github.com/aura-project/aura/blob/80885cca761828fe6f3f2a9c80c04fbc70b35598/src/ChannelServer/Util/GmCommands.cs#L395)).

Your thoughts?

exectails:
I meant only equipped items, and yea, I think text based would be much easier. We have to keep in mind that numbers are hard to remember, and the concept of bitmasks isn't really obvious to everybody, especially newbies.

PR #601, opened 2017-01-16
Added Mentum

eotetb:
Adds the weaponshop dealer Mentum to Nekojima.


eotetb:
Opted to use Mentum's official name so that upgrades work. This of course means that he shows up with a Korean name since his name was never translated in the NA localization.

exectails:
While English names are obviously preferable, maybe we should stick to one of the two for consistency =/ Anyone can still translate those few names in the client after all, we can even provide a DataModder mod.

